PR #1, opened 2021-05-31
Move validation code

dmdashenkov:
In this PR we move the validation codegen from an example repository to this new dedicated repository.

Validation logic is now split into `model` — a language-agnostic base, which assembles validation rules, and `java` — a Java-specific implementation, which generates Java code based on `model`.


dmdashenkov:
@armiol, @alexander-yevsyukov, PTAL.

As there are many changed files, in this PR we will only migrate the codebase from `kanban` and split it into a language-agnostic and a Java-specific part. In further PRs, we're going to add integration tests, along with usage instructions.

model/src/main/java/io/spine/validation/MessageValidationView.java
```diff
@@ -0,0 +1,69 @@
...
+    @Subscribe
+    void on(@External TypeEntered event) {
+        MessageType type = event.getType();
+        builder().setName(type.getName())
+                 .setType(type);
+    }
+
+    @Subscribe
+    void on(SimpleRuleAdded event) {
+        Rule roc = Rule
```
armiol, line 54: I am having hard time imagining why this is `roc` :)

I would understand if that's a `rot` ("rule of thumb"), but with "roc" you got me here and in the next method as well.

model/src/main/java/io/spine/validation/package-info.java
```diff
@@ -0,0 +1,33 @@
...
+ * OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+@CheckReturnValue
```
armiol, line 27: Please document the package.

BTW, do we really want to re-use this package for events, projections etc generated from Proto definitions? I would go for `... .event`, `... .command` — as we usually do for Bounded contexts.

armiol (changes requested):
@dmdashenkov please see my comments.
Also, I am not sure which version this artifact has. I was looking for `version.gradle.kts`, but found none.

dmdashenkov:
@armiol, PTAL.

model/src/main/java/io/spine/validation/MessageValidationRepository.java
```diff
@@ -0,0 +1,51 @@
...
+import org.jetbrains.annotations.NotNull;
+
+import static io.spine.server.route.EventRoute.withId;
+
+/**
+ * A repository for the {@link MessageValidationView}.
+ *
+ * <p>Routes the {@code TypeEntered} events to the view by the type name.
+ */
+class MessageValidationRepository
```
alexander-yevsyukov, line 42: Please document for inheritance or make the class `final`.

model/src/main/java/io/spine/validation/MessageValidationView.java
```diff
@@ -0,0 +1,69 @@
...
+import io.spine.validation.event.CompositeRuleAdded;
+import io.spine.validation.event.SimpleRuleAdded;
+
+/**
+ * A view which accumulates validation data for a message type.
+ *
+ * <p>To add more rules to the message validation, emit {@code SimpleRuleAdded} or
+ * {@code CompositeRuleAdded} events.
+ */
+class MessageValidationView
```
alexander-yevsyukov, line 44: Please document on inheriting from this class, or make it `final`.

model/src/main/java/io/spine/validation/RequiredRulePolicy.java
```diff
@@ -0,0 +1,104 @@
...
+import static io.spine.validation.ComparisonOperator.NOT_EQUAL;
+
+/**
+ * A {@link Policy} which controls whether or not a field should be validated as {@code required}.
+ *
+ * <p>Whenever a field option is discovered, if that option is the {@code required} option, and
+ * the value is {@code true}, and the field type supports such validation, a validation rule
+ * is added. If any of these conditions are not met, nothing happens.
+ */
+public final class RequiredRulePolicy extends Policy<FieldOptionDiscovered> {
```
alexander-yevsyukov, line 54: Why do we need this class `public`?

alexander-yevsyukov (changes requested):

model/src/main/proto/spine/validation/value.proto
```diff
@@ -0,0 +1,154 @@
...
+option java_package = "io.spine.validation";
+option java_outer_classname = "StructProto";
+option java_multiple_files = true;
+
+import "spine/protodata/ast.proto";
+
+// Inspired by the `google/protobuf/struct.proto`.
+//
+// Prefer Google's types when integrating with third-party or when converting to/from JSON.
+// Prefer these types when strict typing matters more than JSON representation.
```
alexander-yevsyukov, line 43: Having two `Prefer` in two consecutive sentences is a bit confusing: What do I have to prefer, what it said in this sentence, or in the previous one?

model/src/main/proto/spine/validation/value.proto
```diff
@@ -0,0 +1,154 @@
...
+import "spine/options.proto";
+
+option (type_url_prefix) = "type.spine.io";
+option java_package = "io.spine.validation";
+option java_outer_classname = "StructProto";
+option java_multiple_files = true;
+
+import "spine/protodata/ast.proto";
+
+// Inspired by the `google/protobuf/struct.proto`.
```
alexander-yevsyukov, line 40: I think we should start the documentation with what the thing does. And then say, what was the source of our inspiration.

model/src/main/proto/spine/validation/rule.proto
```diff
@@ -0,0 +1,164 @@
...
+    GREATER_OR_EQUAL = 6;
+
+    DISTINCT = 7;
+}
+
+// A non-predefined comparison operator.
+message CustomOperator {
+
+    // Describes how the renderer should compare the values.
+    string description = 1;
```
alexander-yevsyukov, line 100: What if we add a field `name` before the `description`? It would be better to have a short form text _and_ a longer one with documentation. It would help formulating both.

java/src/main/java/io/spine/validation/java/JavaValidationRenderer.java
```diff
@@ -0,0 +1,151 @@
...
+    private static final String VIOLATIONS = "violations";
+
+    private @MonotonicNonNull TypeSystem typeSystem;
+
+    public JavaValidationRenderer() {
+        super(ImmutableSet.of(CommonLanguages.getJava()));
+    }
+
+    @Override
+    protected void doRender(SourceSet sources) {
```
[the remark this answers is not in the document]
dmdashenkov: I've added an [issue](https://github.com/SpineEventEngine/ProtoData/issues/17) for this. We'll fix the naming in the next iteration of changes in ProtoData.

model/src/main/proto/spine/validation/rule.proto
dmdashenkov: This type is likely to change. 

At this moment, we simply ignore any custom operators. We also, of course, do not generate them.

In future, we might need some kind of a way to inject renderers with non-predefined operations. I assume we might simply put code expressions in target languages in there.

In this PR, I'd like to leave this type alone and see how it goes when we add more validation rules into the mix. The structure of a `CustomOperator` is unknown for now.

dmdashenkov:
@alexander-yevsyukov, PTAL.

model/src/main/java/io/spine/validation/RequiredRulePolicy.java
```diff
@@ -0,0 +1,104 @@
...
+ * is added. If any of these conditions are not met, nothing happens.
+ */
+final class RequiredRulePolicy extends Policy<FieldOptionDiscovered> {
+
+    @NotNull
+    @Override
+    @React
+    public EitherOf2<SimpleRuleAdded, Nothing> whenever(@External FieldOptionDiscovered event) {
+        Option option = event.getOption();
+        if (isRequired(option)) {
```
alexander-yevsyukov, line 61: Please invert the `if` condition. It would reduce the nesting.

model/src/main/java/io/spine/validation/RequiredRulePolicy.java
```diff
@@ -0,0 +1,104 @@
...
+                            "Unknown file `%s`.", event.getFile()
+                                                       .getValue()
+                    ));
+            MessageType type = file.getTypeMap()
+                                   .get(typeUrl(event.getType()));
+            Field field = type.getFieldList()
+                              .stream()
+                              .filter(f -> f.getName().equals(event.getField()))
+                              .findAny()
+                              .orElseThrow(() -> newIllegalArgumentException(
```
alexander-yevsyukov, line 74: Please extract creation on this and previous exception. It would make this complex method more readable.

model/src/main/java/io/spine/validation/RequiredRulePolicy.java
```diff
@@ -0,0 +1,104 @@
...
+        }
+        return EitherOf2.withB(nothing());
+    }
+
+    private static SimpleRuleAdded requiredRule(Field field) {
+        Value defaultValue = DefaultValue.forField(field);
+        @SuppressWarnings({"DuplicateStringLiteralInspection", "RedundantSuppression"})
+            // Duplication in generated code.
+        SimpleRule rule = SimpleRule.newBuilder()
+                                    .setErrorMessage("Field must be set.")
```
alexander-yevsyukov, line 87: Can we wrap this block to the left, please.

model/src/main/java/io/spine/validation/RequiredRulePolicy.java
```diff
@@ -0,0 +1,104 @@
...
+                                    .setField(field)
+                                    .setSign(NOT_EQUAL)
+                                    .setOtherValue(defaultValue)
+                                    .vBuild();
+        return SimpleRuleAdded.newBuilder()
+                              .setType(field.getDeclaringType())
```
alexander-yevsyukov, line 93: Can we wrap this block to the left, please.

model/src/main/java/io/spine/validation/RequiredRulePolicy.java
```diff
@@ -0,0 +1,104 @@
...
+/**
+ * A {@link Policy} which controls whether or not a field should be validated as {@code required}.
+ *
+ * <p>Whenever a field option is discovered, if that option is the {@code required} option, and
+ * the value is {@code true}, and the field type supports such validation, a validation rule
+ * is added. If any of these conditions are not met, nothing happens.
+ */
+final class RequiredRulePolicy extends Policy<FieldOptionDiscovered> {
+
+    @NotNull
```
alexander-yevsyukov, line 56: Can we do without this annotation?

model/src/main/java/io/spine/validation/SpineOptionsProvider.java
```diff
@@ -0,0 +1,47 @@
...
+ * An {@link OptionsProvider} which registers Spine options from {@code options.proto}.
+ *
+ * <p>By registering Spine options we allow ProtoData to access them when reading the descriptors.
+ * Otherwise, validation options would be unrecognized and validation would not work.
+ */
+@SuppressWarnings("unused") // Accessed via reflection by ProtoData.
+public final class SpineOptionsProvider implements OptionsProvider {
+
+    @Override
+    public void dumpTo(@NotNull ExtensionRegistry registry) {
```
alexander-yevsyukov, line 44: Please use annotations from Checker Framework.

alexander-yevsyukov (changes requested):
Please see my comments.

alexander-yevsyukov (approved):
LGTM

model/src/main/java/io/spine/validation/DefaultValue.java
```diff
@@ -0,0 +1,132 @@
...
+import static com.google.common.base.Preconditions.checkNotNull;
+import static io.spine.protodata.PrimitiveType.PT_UNKNOWN;
+import static io.spine.protodata.PrimitiveType.TYPE_BYTES;
+import static io.spine.protodata.PrimitiveType.TYPE_STRING;
+import static io.spine.protodata.PrimitiveType.UNRECOGNIZED;
+import static io.spine.util.Exceptions.newIllegalArgumentException;
+import static io.spine.util.Exceptions.newIllegalStateException;
+
+/**
+ * A factory of {@link Value}s representing default states of Protobuf message fields.
```
armiol, line 47: As discussed vocally, let's expand this documentation piece.

In particular, we need to describe the difference between this factory and `io.spine.protobuf.Messages` (which on the surface looks more promising, equipped with some caching strategy).

armiol (changes requested):
@dmdashenkov please see my comment, as per our discussion.

dmdashenkov:
@armiol, PTAL.

I've renamed the `DefaultValue` class to `NotSetValue`. After a re-evaluation, it seems a bit more appropriate to its purpose.
I don't like "Unset" as it reads as if there used to be a value but not anymore.
Let's discuss the changes in person if necessary.

armiol (approved):
@dmdashenkov LGTM.

PR #2, opened 2021-07-19
Integration tests

dmdashenkov:
In this PR we add integration tests for the validation code generation.
We introduce two modules, one of which acts as an extension to the standard validation library, and the other declared the validation rules and launches ProtoData with all the validation components.

We also add the support for `(min)`, `(max)`, and `(range)` options.

Also, starting with this PR, we publish artifacts to the Maven repository associated with this GitHub repository. Previously, we attempted publishing into ProtoData's Maven repository, but failed due to the lack of permissions in the standard GH credentials. This hints at the intention of GH Packages designers to separate artifacts repo-wise. We will try that for now and see if it's handy enough.


dmdashenkov:
@armiol, PTAL.

java/src/main/kotlin/io/spine/validation/java/GenerationContext.kt
```diff
@@ -0,0 +1,126 @@
...
+     */
+    val typeSystem: TypeSystem,
+
+    /**
+     * The type of the validated message.
+     */
+    val declaringType: TypeName,
+
+    /**
+     * A reference to the mutable violations list, which accumulates all the constraint violations.
```
armiol, line 72: I would describe it in more details. I assume now that this is about a variable name? But I am not sure about that. Therefore, let's make it more clear.

java/src/main/kotlin/io/spine/validation/java/GenerationContext.kt
```diff
@@ -0,0 +1,126 @@
...
+    /**
+     * Finds the field by the given name in the validated type.
+     *
+     * @throws IllegalArgumentException if there is no such field
+     */
+    fun lookupField(name: FieldName): Field =
+        querying.lookUpField(protoFile, declaringType, name)
+}
+
+private fun Querying.lookUpField(file: FilePath, type: TypeName, field: FieldName): Field {
```
armiol, line 108: Let's settle on either `lookUp` (like here) or `lookup` (a few lines above).

java/src/main/kotlin/io/spine/validation/java/GenerationContext.kt
```diff
@@ -0,0 +1,126 @@
...
+    val protoFile = select<ProtobufSourceFile>().withId(file).orElseThrow {
+        IllegalArgumentException("Unknown file: `${file.value}`.")
+    }
+    val messageType = protoFile.typeMap[type.typeUrl()]
+        ?: throw IllegalArgumentException("Unknown type: `${type.typeUrl()}`.")
+    return messageType.findField(field)
+        ?: throw IllegalArgumentException("Unknown field: `${type.typeUrl()}.${field.value}`.")
+}
+
+private fun MessageType.findField(name: FieldName): Field? {
```
armiol, line 118: Can we name all of them similarly? I mean, can we make this `lookup...` (or `lookUp...`)  as well?

java/src/main/kotlin/io/spine/validation/java/JavaCodeGenerator.kt
```diff
@@ -89,67 +88,74 @@ private val BOOLEAN_OPS = mapOf(
...
-        val field = rule.field
         val type = field.type
         val signs = if (type.kindCase == PRIMITIVE) {
             PRIMITIVE_COMPARISON_OPS
         } else {
             OBJECT_COMPARISON_OPS
         }
-        val compare = signs[rule.sign]!!
+        val compare = signs[rule.sign]
+            ?: throw IllegalStateException("Cannot render ${rule.sign} in Java for type `$type`.")
```
armiol, line 144: When reading this in a stacktrace, I'd like to have at least some reasoning :)

model/src/main/kotlin/io/spine/validation/ErrorMessage.kt
```diff
@@ -27,21 +27,29 @@
...
+import io.spine.validation.Placeholder.OTHER
+import io.spine.validation.Placeholder.RIGHT
+import io.spine.validation.Placeholder.VALUE
 
 /**
  * A human-readable error message, describing a validation constraint violation.
+ *
+ * The error message can contain dynamic references to values in the generated code.
+ * If such references are present, they are inserted into the message via the string plus (`+`)
+ * operators.
```
armiol, line 44: I would end this sentence like this:

```
... into the message via the plus (`+`) operator.
```
and make it fit the previous line.

model/src/main/kotlin/io/spine/validation/ErrorMessage.kt
```diff
@@ -71,27 +84,68 @@ private constructor(private val value: String) {
...
+     * In a composite validation rule, the right condition.
+     */
+    RIGHT,
+
+    /**
+     * In a composite validation rule, the boolean operator which joins the two conditions.
+     *
+     * @see LogicalOperator
+     */
+    OPERATION;
```
armiol, line 144: The description and `@see` says "operator". I would name this one `OPERATOR` as well.

model/src/main/kotlin/io/spine/validation/RangeNotation.kt
```diff
@@ -0,0 +1,111 @@
...
+
+import com.google.common.collect.Range
+import io.spine.validate.ComparableNumber
+import java.lang.IllegalStateException
+import java.util.regex.Pattern
+
+
+/**
+ * Transforms a string value defined in a field declaration into a [Range] of
+ * [ComparableNumber]s.
```
armiol, line 37: Let's not leave it alone on this line.

model/src/main/proto/spine/validation/rule.proto
```diff
@@ -132,13 +132,24 @@ message CompositeRule {
...
     string error_message = 4;
+
+    // If this rule describes several constraints of the same field, this is the name of the field.
+    //
+    // It is only used for diagnostic purposes. The constraints themselves are always validated
+    // according to the rules `left` and `right`.
+    //
+    // If `left` and `right` do not describe constraints of the same field, this value is empty.
+    //
+    protodata.FieldName common_field = 5;
```
armiol, line 146: I would never guess what this field is by its name.

Can we rename it to `constrained_field`? With the same docs.

armiol (changes requested):
@dmdashenkov LGTM in general. Please see my comments.

model/src/main/proto/spine/validation/rule.proto
dmdashenkov: Renamed to `field`, just like in `SimpleRule`.

The intention was to show that the field is common across `left` and `right` rules. But, if the name confuses the reader rather than helps, I'd prefer a shorter name.

dmdashenkov:
@armiol, PTAL again.

armiol (approved):

PR #4, opened 2021-07-27
More rules

dmdashenkov:
In this PR we implement more validation rules:
 - `(distinct)` for collections forces a collection not to have duplicate elements;
 - `(validate)` for a message field forces the message to be validated as a part of the parent message validation.

We also introduce the mechanism for rendering custom validation operators.
Sure enough, the mechanism requires the API user to write their own `Renderer`. However, we prepare the environment for such a renderer, as well as provide examples of how to generate code for custom operators.


dmdashenkov:
@armiol, PTAL.

java/src/main/java/io/spine/validation/java/JavaValidationRenderer.java
```diff
@@ -142,17 +190,34 @@ private CodeBlock generateValidationCode(MessageValidation validation,
...
+    private static CodeBlock generateValidationError() {
         CodeBlock.Builder code = CodeBlock.builder();
         code.beginControlFlow("if (!$N.isEmpty())", VIOLATIONS);
-        code.addStatement("throw new $T($N)", ValidationException.class, VIOLATIONS);
+        MethodCall errorBuilder = new ClassName(ValidationError.class)
+                .newBuilder()
+                .chainAddAll("constraint_violation", new Literal(VIOLATIONS))
+                .chainBuild();
+        MethodCall newOptional = new ClassName(Optional.class)
+                .call("of", newArrayList(errorBuilder), newArrayList());
```
armiol, line 215: If I remember correctly, `newArrayList()` is deprecated for Java starting 7th. Please review.

model/src/main/java/io/spine/validation/SimpleRules.java
```diff
@@ -0,0 +1,76 @@
...
+     *         the target field
+     * @param customFeature
+     *         the feature message describing the custom operator
+     * @param description
+     *         the human-readable text description of the feature
+     * @param errorMessage
+     *         the error message for the case of violation
+     * @return a new rule
+     */
+    static SimpleRule withCustom(FieldName field,
```
armiol, line 58: I would check the parameters for staying in expected boundaries (non-null, non-empty, etc).

model/src/main/proto/spine/validation/rule.proto
```diff
@@ -68,6 +68,12 @@ message SimpleRule {
     // when the error is constructed.
     //
     string error_message = 5;
+
+    // If `true`, this rule is ignored when the field is not set.
+    //
+    // This does not apply to simple rules which are parts of a `CompositeRule`.
+    //
+    bool ignore_if_not_set = 6;
```
armiol, line 76: Looking at it as at a `Message` property I would consider naming it `ignored_if_unset`.

"ignore_..." looks more like a command to me.

model/src/main/proto/spine/validation/standard_features.proto
```diff
@@ -0,0 +1,64 @@
...
+// These are validation features which do not fit into the simple rule formula:
+// ```
+// msg.field <=> other_value
+// ```
+//
+// All validation renderers should generate code for these features in order to comply with
+// the validation protocol.
+
+// The associated collection field, a list or a map, must only contain distinct elements,
+// i.e. repetition is not allowed.
```
armiol, line 51: Please clarify which elements of `map` you mean.

armiol (approved):

armiol:
@dmdashenkov LGTM, but please see my comments.

PR #6, opened 2021-08-18
Distribute rules to elements of repeated fields

dmdashenkov:
In this PR we enable list and map Protobuf fields to have the same validation rules as do singular fields.
This includes distinguishing between rules that are applied to the whole collection and rules that are applied to individual elements.

### DIFF
--- a/java/src/main/kotlin/io/spine/validation/java/CustomRuleGenerators.kt
+++ b/java/src/main/kotlin/io/spine/validation/java/CustomRuleGenerators.kt
@@ -105,7 +105,7 @@ private class ValidateGenerator(ctx: GenerationContext) : SimpleRuleGenerator(ct
     }
 
     override fun condition(): Expression =
-        MethodCall(violationsVariable, "isPresent")
+        Literal("!" + MethodCall(violationsVariable, "isPresent"))
 
 
     override fun createViolation(): CodeBlock {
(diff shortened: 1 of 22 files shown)


dmdashenkov:
@armiol, PTAL.

java/src/main/kotlin/io/spine/validation/java/CustomRuleGenerators.kt
```diff
@@ -71,51 +76,60 @@ private class DistinctGenerator(ctx: GenerationContext) : SimpleRuleGenerator(ct
...
+private class ValidateGenerator(ctx: GenerationContext) : SimpleRuleGenerator(ctx) {
+
+    private val violationsVariable =
+        Literal("generated_validationError_${ctx.fieldFromSimpleRule!!.name.value}")
+
+    init {
+        val field = ctx.fieldFromSimpleRule!!
+        val fieldType = field.type
+        check(fieldType.hasMessage()) {
+            "(validate) only supports message types but field " +
```
armiol, line 91: Can we say 
```
... only supports `Message` types ...
```?

File: java/src/main/kotlin/io/spine/validation/java/GenerationContext.kt [hunk 2]
```diff
@@ -95,10 +100,24 @@ internal data class GenerationContext(
             null
         }
 
-    /**
-     * Obtains the same context but with the given validation [rule].
-     */
-    fun withRule(rule: Rule): GenerationContext = copy(rule = rule)
+    val fieldOrElement: Expression?
```
armiol, line 103: Let's document the `val`s.

armiol (approved):
@dmdashenkov LGTM with a few comments to address.

PR #9, opened 2021-09-06
Custom error messages

dmdashenkov:
Until now, the only possible error message that came with a constraint violation was a hard-coded into the validation model message.

In this PR, we start supporting custom error messages defined by the user alongside the message definition.
Most options provide a special `msg_format` string, that can be used to specify a custom error message.
Such a format supports insertion points:
 - `{value}` — for the actual value of the validated field;
 - `{other}` — for the value to which the value of the field is compared.

Printf-style placeholders are not supported anymore. This change will be reflected in the options doc as a part of the upcoming migration, which will move the options from `base` into this repository.

Because printf placeholders are not supported, default error messages for some of the options are not obtained from the `(default_message)` option, but are instead hardcoded in the model. This as well will be changed as a part of the option migration.


dmdashenkov:
@alexander-yevsyukov, PTAL.

model/src/main/java/io/spine/validation/MaxPolicy.java
```diff
@@ -46,18 +45,17 @@
...
-            return EitherOf2.withB(nothing());
-        }
         NumberRules rules = NumberRules.from(option);
         SimpleRule rule = rules.maxRule(event.getField());
-        return EitherOf2.withA(SimpleRuleAdded
-                                       .newBuilder()
-                                       .setType(event.getType())
-                                       .setRule(rule)
-                                       .build()
+        return new Just<>(SimpleRuleAdded
```
alexander-yevsyukov, line 54: Can we have a static factory called `just()` in the `Just` type so that we can write simply this:

```java
return just(SimpleRuleAdded.newBuilder()...);
```
?

model/src/main/java/io/spine/validation/ValidatePolicy.java
```diff
@@ -57,37 +54,47 @@
...
 
     @Override
     @React
-    public EitherOf2<SimpleRuleAdded, Nothing> whenever(@External FieldOptionDiscovered event) {
-        Option option = event.getOption();
-        if (!is(option, validate) || !unpack(option.getValue(), BoolValue.class).getValue()) {
-            return EitherOf2.withB(nothing());
+    public EitherOf2<SimpleRuleAdded, Nothing> whenever(@External FieldExited event) {
+        FieldId id = FieldId
+                .newBuilder()
```
alexander-yevsyukov, line 63: Please have `.newBuilder() ` on the previous line. IDEA already can align things nicely, and we have options set for this.

model/src/main/java/io/spine/validation/ValidatePolicy.java
```diff
@@ -57,37 +54,47 @@
...
+            SimpleRule rule = SimpleRules.withCustom(
+                    event.getField(),
+                    RecursiveValidation.getDefaultInstance(),
+                    "Message field is validated by its validation rules. " +
+                            "If the field is invalid, the container message is invalid as well.",
+                    field.get()
+                         .getErrorMessage(),
+                    true);
+            return EitherOf2.withA(SimpleRuleAdded
+                                           .newBuilder()
```
alexander-yevsyukov, line 83: How about this layout:

```java
return EitherOf2.withA(
    SimpleRuleAdded.newBuilder()
        .setType(event.getType())
        .setRule(rule)
        .build()
    );
}

File: model/src/main/proto/spine/validation/field_views.proto [hunk 4]
```diff
@@ -0,0 +1,51 @@
...
+
+    // Message appearing when this required field is not set.
+    string error_message = 2;
+
+    // `true` if the field is marked with `(required) = true`, `false` otherwise.
+    bool required = 3;
+}
+
+// A view on a field that is marked with `(validate)`.
+//
```
alexander-yevsyukov, line 30: Please use an empty line only between paragraphs here and above.

alexander-yevsyukov (approved):
LGTM with one suggestion as to static factory method, and minor code layout notes. Please consider before merging.

model/src/main/java/io/spine/validation/MaxPolicy.java
dmdashenkov: Will add to ProtoData in the next iteration. Here is the [issue](https://github.com/SpineEventEngine/ProtoData/issues/39).

PR #14, opened 2021-11-15
`When` constraint

dmdashenkov:
In this PR we introduce support for the `(when)` constraint. Now the generated code will check that a value (a `google.protobuf.Timestamp` or anything that can be compared to a `Timestamp`) is in the past/future.

### DIFF
--- a/test/consumer/src/test/java/io/spine/validation/test/ValidationTest.java
+++ b/test/consumer/src/test/java/io/spine/validation/test/ValidationTest.java
@@ -320,7 +320,7 @@ class ValidationTest {
     }
 
     @Nested
-    @DisplayName("reflect the `(when)` rule"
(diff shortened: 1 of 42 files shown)


dmdashenkov:
@armiol, PTAL.

test/consumer/src/test/java/io/spine/validation/test/ValidationTest.java
```diff
@@ -317,6 +319,29 @@ void pass() {
         }
     }
 
+    @Nested
+    @DisplayName("reflect the (when) rule")
```
armiol, line 323: Maybe
```
`(when)`
```
?

armiol (approved):
@dmdashenkov LGTM. However, I will ask you for an assistance to fully understand how it is going to work.

PR #15, opened 2021-11-18
Delegation fixes

dmdashenkov:
In this PR we fix small issues discovered when applying the new validation lib to `mc-java`.
We will later move tests from `mc-java` into `validation` (in a separate PR).


dmdashenkov:
@armiol, PTAL.

armiol (approved):
I also wonder whether we had any tests that did not fail, or if we plan to introduce any tests in the future to cover these changes. Let's discuss later today.

PR #17, opened 2021-12-08
Required IDs and more informative error messages

dmdashenkov:
In this PR we add support for the required ID rules.

When a special kind of message is defined, an event, a command, a rejection, or an entity state, its first field must be the ID field. The ID fields are required by default (unless `(required) = false`).

Also, in this PR we tighten the rule for required collections. Previously, we only checked the collection to not be empty. Now, we also check the elements to be non-empty (non-default). This does not apply to primitive types that have no concept of an empty value, such as numbers and `bool`s.


model/src/main/java/io/spine/validation/RequiredIdPatternPolicy.java
```diff
@@ -0,0 +1,93 @@
...
+
+import static io.spine.validation.SourceFiles.findFirstField;
+
+/**
+ * A policy that marks ID fields in entity state messages and signal messages as required.
+ *
+ * <p>The messages are discovered via the file patterns, specified in {@link ValidationConfig}.
+ * If ProtoData runs with no config, this policy never produces any validation rules.
+ */
+final class RequiredIdPatternPolicy extends RequiredIdPolicy {
```
armiol, line 48: Let's describe the difference between this policy and `RequiredIdOptionPolicy`. As discussed, we need to guide the potential users on which to apply and when.

model/src/main/proto/spine/validation/config.proto
```diff
@@ -0,0 +1,45 @@
...
+option java_package = "io.spine.validation";
+option java_outer_classname = "ConfigProto";
+option java_multiple_files = true;
+
+import "spine/tools/mc/java/config.proto";
+
+// The container type for the validation config.
+//
+// ProtoData supports multiple configurations merges into a single object. In order for
+// the Validation library to be satisfied with the config, it mast always have
```
armiol, line 17: `must`.

armiol (changes requested):
@dmdashenkov LGTM in general. Please see my comments.

model/src/main/java/io/spine/validation/RequiredIdPatternPolicy.java
```diff
@@ -0,0 +1,93 @@
...
+                                    .anyMatch(pattern -> matches(file, pattern));
+        if (!match) {
+            return withNothing();
+        }
+        TypeName type = event.getType();
+        Field field = findFirstField(type, file, this);
+        return withField(field);
+    }
+
+    private static ImmutableList<FilePattern> allPatterns(MessageMakers markers) {
```
alexander-yevsyukov, line 70: Please consider transforming this conversion to Kotlin extension for `MessageMakers`.

model/src/main/java/io/spine/validation/ValidationPolicy.java
```diff
@@ -0,0 +1,69 @@
...
+/**
+ * A policy that reacts to an event with a {@link RuleAdded} event.
+ *
+ * <p>May ignore an event and return {@code Nothing} if necessary.
+ *
+ * @param <E>
+ *         the type of the event to react to
+ */
+public abstract class ValidationPolicy<E extends EventMessage>
+        extends Policy<E> {
```
alexander-yevsyukov, line 45: Please put `extends` on the previous line. It should fit.

model/src/main/java/io/spine/validation/Rules.java
```diff
@@ -0,0 +1,60 @@
...
+    /**
+     * Prevents the utility class instantiation.
+     */
+    private Rules() {
+    }
+
+    /**
+     * Converts the given {@code rule} to an event.
+     */
+    static RuleAdded toEvent(Rule rule, TypeName type) {
```
alexander-yevsyukov, line 47: Please consider transforming this into a Kotlin extension for the `Rule` type.

Please also document the `type` parameter.

model/src/main/java/io/spine/validation/RequiredRule.java
```diff
@@ -48,32 +53,89 @@ private RequiredRule() {
...
+    }
+
+    private static SimpleRule rule(Field field,
+                                   Value value,
+                                   String errorMessage,
+                                   String defaultErrorMessage,
+                                   boolean distibute) {
+        String msg = errorMessage.isEmpty() ? defaultErrorMessage : errorMessage;
+        return SimpleRule
                 .newBuilder()
```
alexander-yevsyukov, line 102: Please put `newBuilder()` on the previous line.

model/src/main/java/io/spine/validation/RequiredRule.java
```diff
@@ -48,32 +53,89 @@ private RequiredRule() {
...
-                "Field `%s.%s` of type `%s` does not support `(required)` validation.",
-                typeUrl, fieldName, type
-        ));
+    private static Rule wrap(SimpleRule r) {
+        return Rule.newBuilder()
+                .setSimple(r)
+                .build();
+    }
+
+    private static Rule wrap(CompositeRule r) {
```
alexander-yevsyukov, line 117: Can these two methods be a part of a mixin interface common for `SimpleRule` and `CompositeRule`?

model/src/main/java/io/spine/validation/RequiredPolicy.java
```diff
@@ -65,15 +67,24 @@
...
+        return Rules.toEvent(rule, declaration.getDeclaringType());
+    }
+
+    private static IllegalStateException doesNotSupportRequired(Field field) {
+        String fieldName = field.getName()
+                                .getValue();
+        String typeUrl = typeUrl(field.getDeclaringType());
+        PrimitiveType type = field.getType()
+                                  .getPrimitive();
+        return new IllegalStateException(format(
```
alexander-yevsyukov, line 85: We have `Exceptions.newIllegalStateException()` which does exactly this. Why not use it?

alexander-yevsyukov (approved):
LGTM, with a few suggestions to consider before merging.

dmdashenkov:
@armiol, PTAL.

armiol (approved):

PR #31, opened 2022-08-29
Avoid the runtime dependencies onto `validation-runtime`

armiol:
This changeset updates the generated validation code to use `CollectionsConverter` [introduced](https://github.com/SpineEventEngine/base/pull/723) to `spine-base`.

The idea is to get rid of `validation-runtime` artifact when compiling the Java files enhanced with the generated validation code. It is going to be enough that the generated code uses `spine-base` artifact already.

The `validation-runtime` module is left intact in this project to avoid any compatibility issues, as there are some complex Gradle dependency relations between `validation`, `mc-java` and `ProtoData` libraries, which are addressed partially by forcing some artifact versions. The `validation-runtime` module is now marked deprecated and will be removed in the following changesets.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -77,8 +77,7 @@ spinePublishing {
         ":proto:configuration",
         ":proto:context",
         "java",
-        "model",
-        "runtime"
+        "model"
     )
     destinations = with(PublishingRepos) {
         setOf(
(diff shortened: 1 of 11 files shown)


armiol:
@alexander-yevsyukov PTAL.

@dmdashenkov FYI.

PR #41, opened 2022-09-28
Fat JAR bundle

armiol:
As a next step in continuous effort to run ProtoData in isolation (meaning, minimum of external dependencies, and no conflicts with Spine libraries-under-development), in this changeset a new fat JAR is now being assembled and published.

From now on, `spine-validation-java-bundle:all` artifact is published along with previously published artifacts. It is a no-dependency JAR containing the codegen plugin for Java code on top of ProtoData. 

In subsequent PRs, this JAR will be used by `modelCompiler` upon configuring the ProtoData classpath instead of previously used `spine-validation-java`, which unfortunately had a long list of dependencies.

### DIFF
--- a/java-bundle/build.gradle.kts
+++ b/java-bundle/build.gradle.kts
@@ -1,3 +1,5 @@
+import io.spine.internal.gradle.publish.SpinePublishing
+
 /*
  * Copyright 2022, TeamDev. All rights reserved.
  *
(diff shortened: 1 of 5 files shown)


armiol:
@dmdashenkov @alexander-yevsyukov PTAL.

dmdashenkov (approved):
LGTM with a single comment.

PR #47, opened 2022-10-05
Update Protobuf, fix publishing

armiol:
This changeset performs two operations:

1. Updates the Protobuf library version since the one previously used has some recently discovered vulnerabilities. Also, it migrates the project to the latest `base` library, which also has its Protobuf dependency updated.

2. Excludes all transitive dependencies onto stale `io.spine:spine-validate` artifact, which are somehow brought via the project dependencies. It caused this PR to fail due to the version conflicts with this artifact.

3. Fixes (and changes) the publishing of Validation modules. There is now two fat JAR artifacts published:

* `io.spine.validation:spine-validation-java-extensions` — ProtoData extensions allowing to generate the Validation code in Java;
* `io.spine.validation:spine-validation-java-runtime` which is a runtime library required for the generated Java code to work.

Please note that both of the published fat JARs have NO classifier. It has been removed for simplicity.

The library version has been set to `2.0.0-SNAPSHOT.29`.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -80,6 +80,7 @@ spinePublishing {
         ":proto:context",
         "java",
         "java-bundle",
+        "java-runtime-bundle",
         "model"
     )
     destinations = with(PublishingRepos) {
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -170,6 +170,9 @@ subprojects {
                     Dokka.BasePlugin.lib
                 )
             }
+
+            // Exclude all transitive dependencies onto the recently moved artifact.
+            exclude("io.spine", "spine-validate")
         }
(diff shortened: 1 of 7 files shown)


armiol:
@alexander-yevsyukov PTAL.

Hopefully, once this is merged `mc-java` will become buildable again. I will let you know once I am done with `mc-java`'s branch.

alexander-yevsyukov:
@armiol, can we have the 2nd artifact name shorter, like `spine-validation-java-runtime`?

alexander-yevsyukov (changes requested):
LGTM, but the runtime bundle name needs to avoid the `extensions` suffix. It's a bit confusing.

PR #53, opened 2022-10-06
Deprecate `vBuild()`

alexander-yevsyukov:
This PR deprecates `ValidatingBuilder.vBuild()` in favor of `build()` which now is also annotated `@Validated` since Validation already can patch the standard Prototuf-generated code with our features.

Setting `@Validated` annotation on generated `build()` methods would be done in a separate PR.


armiol (approved):
@alexander-yevsyukov LGTM.

PR #54, opened 2022-10-07
Make generated messages implement `ValidatableMessage`

alexander-yevsyukov:
This PR teaches Validation to make generated code of message types to implement `ValidatableMessage` interface.

This interface used to be named `MessageWithConstraints` and had `validate()` method which returned a list of `ValidationConstraint`. Now it returns `Optional<ValidationError>` as in the recently generated code.

Other notable changes:
  * The code which used to work with `List<String>` now uses the `Text` type recently introduced in `base`.
  * The code was refactored to introduce abstract class `BuilderInsertionPoint`.
  * The descendants of this class named `BuildMethodReturnTypeAnnotation` and `BuildPartialMethodReturnTypeAnnotation` were also introduced, but could not be fully implemented because of [this limitation](https://github.com/SpineEventEngine/ProtoData/issues/84) in ProtoData.
  * The `JavaValidationRenderer` was refactored extracting method objects to make the whole process easier to understand.
  * The `Validate` class was renamed to `ValidateBeforeReturn` to better reflect the nature of the insertion point.


alexander-yevsyukov:
@armiol, @dmdashenkov,
  
In relation to the `Text` type, we probably need even a type with such or similar name introduced at the level of ProtoData or even as a separate small library under the package `io.spine.text`.

We definitely have a generic subdomain which deals with text blocks, lines of text, etc. Passing `ImmutableList<String>` was OK for the start, but 1) it's too many characters to read, and 2) let's not limit ourselves to generating the code only in Java.

We may also benefit from having protobuf-based type `Text` in other projects.

Let's discuss this when time permits.

java/src/main/java/io/spine/validation/java/ValidationCode.java
```diff
@@ -0,0 +1,169 @@
...
+    }
+
+    private void annotateBuildPartialMethod() {
+        var buildPartialMethod = new BuildPartialReturnTypeAnnotation(messageType);
+        sourceFile.at(buildPartialMethod)
+                  .add(annotation(NonValidated.class));
+    }
+
+    private static String annotation(Class<? extends Annotation> annotationClass) {
+        // Add space before `@` so that when the type is fully qualified, the annotation
```
armiol, line 165: Please move it to the method-level docs. It looks an `implNote` to me.

version.gradle.kts
```diff
@@ -24,10 +24,10 @@
...
  */
 
-val spineBaseVersion by extra("2.0.0-SNAPSHOT.108")
+val spineBaseVersion by extra("2.0.0-SNAPSHOT.112")
 val spineToolBaseVersion by extra("2.0.0-SNAPSHOT.109")
 val spineTimeVersion by extra("2.0.0-SNAPSHOT.108")
-val spineServerVersion by extra("2.0.0-SNAPSHOT.111")
+val spineServerVersion by extra("2.0.0-SNAPSHOT.112")
 val protoDataVersion by extra("0.2.16")
 val mcJavaVersion by extra("2.0.0-SNAPSHOT.100")
```
armiol, line 32: I am not sure, but it looks the own library version is not increased below. Please check.

armiol (approved):
@alexander-yevsyukov LGTM with some comments to consider.

java/src/main/java/io/spine/validation/java/ParsedSources.java
```diff
@@ -40,16 +41,16 @@ final class ParsedSources {
...
      * Cached results of parsing the Java source code.
      */
-    private final LoadingCache<String, JavaSource<?>> cache =
+    private final LoadingCache<Text, JavaSource<?>> cache =
             CacheBuilder.newBuilder()
                     .maximumSize(300)
                     .build(loader());
 
-    private static CacheLoader<String, JavaSource<?>> loader() {
+    private static CacheLoader<Text, JavaSource<?>> loader() {
```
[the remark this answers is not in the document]
alexander-yevsyukov: I did it, when introduced in `base`. Though, I do think we need to have it as a proto type with appropriate mixin soon.

version.gradle.kts
alexander-yevsyukov: Well-spotted again. Strangely,  `IncrementGuard` is applied to all subprojects at the root `build` script, but it didn't find the conflict.

PR #65, opened 2022-10-27
Bump `base` etc.

alexander-yevsyukov:
This PR:
  * Updates `base` and other dependencies on Spine modules and those from `config`.
  * Removes extra version build properties.
  * Applies proto Kotlin DSL to `TypeSystemSpec` test suite.

There are some `MakeTestFilterHappyTest.java` files in this change set. They are needed because of [this issue](https://github.com/SpineEventEngine/config/issues/421).


alexander-yevsyukov:
Pardon, does not see Kotlin DSL at CI again.

alexander-yevsyukov:
@armiol, @dmdashenkov, I had to rollback usage of Kotlin proto-DSL because it breaks compilation at CI (and locally, if you make a pause between `clean` and `build`).

This time it doesn't find DSL code which is part of ProtoData `compiler` artifact. The code is there. Why it's not visible to the Kotlin compiler in some circumstances is a mystery to me.

Anyway, PTAL.

armiol:
@alexander-yevsyukov this whole story with DSL is weird to me. Not much changed in our codebase, and yet such a thing was never observed on CI previously. Can it be caused by some unfinished business in the Gradle nightly build that we started using everywhere recently?

Just out of curiosity, let's also try the latest nightly build (see https://gradle.org/release-nightly/).

armiol (approved):

armiol:
@alexander-yevsyukov OK, I see that you are using the latest nightly build already. Maybe, that's what breaks stuff? Let's try rolling back to the previously used nightly build (the one we started to use first). Also, if we have any Gradle cache enabled at CI, let's disable it in the scope of GH Actions pipeline.

alexander-yevsyukov:
> OK, I see that you are using the latest nightly build already.

I switched to this version in the hope that it's fixed by the latest Gradle code. I can switch back, but I've already tried many different nightbuilds of Gradle when trying to solve this issue in ProtoData. It didn't work.

I guess I need to have a closer look at how we filter `KotlinCompile` task input in ProtoData. There should be something wrong there.

PR #66, opened 2022-10-29
Bump ProtoData, apply detekt

alexander-yevsyukov:
This PR:
  * Updates dependency on ProtoData and other Spine modules.
  * Applies detekt Gradle plugin and addresses reported errors.
  * Removes detekt GitHub workflow as redundant.


build.gradle.kts
```diff
@@ -61,12 +61,8 @@ buildscript {
 plugins {
     `java-library`
     idea
-
-    val protobuf = io.spine.internal.dependency.Protobuf.GradlePlugin
```
alexander-yevsyukov, line 65: We cannot do it any more after detekt is added to classpath in `buildSrc/build.gradle.kts`.

alexander-yevsyukov: It gives compile error, saying `io` is uknown.

armiol (approved):

PR #70, opened 2022-11-01
Move on

alexander-yevsyukov:


armiol (approved):
I don't see how you actually fix the conflict... but as long as you have the idea, LGTM.

alexander-yevsyukov:
> I don't see how you actually fix the conflict... but as long as you have the idea, LGTM.

We have another module already, which is published with the same coordinates. It conflict caused error. 
A different name for the bundle publishing should solve this issue. At least, it works without problems to `mavenLocal()`.

PR #71, opened 2022-11-01
Fix publishing again

alexander-yevsyukov:
This PR attempts to fix the publishing problem, which wasn't fixed by the previous PR.


alexander-yevsyukov:
@armiol, you were right... :)

armiol (approved):
Now I think the bundle won't publish. But let's try.

PR #73, opened 2022-11-02
Extract `fatJar` publishing

alexander-yevsyukov:
This PR:
  * Extracts publishing of `fatJar` artifact into `fat-jar` scripting plugin.
  * Extends `spinePublishing` with the ability to handle custom publications, introducing `modulesWithCustomPublishing` property.
  * Applies Gradle Shadow plugin at `buildScript/.../build.gradle.kts` level so that `fat-jar` script plugin is possible.

The version is set to `2.0.0-SNAPSHOT.60` to jump over `5x` artifacts that were not successfully published because of errors in the previous PRs.

Once this PR is merged and new publishing is fully tested, the extensions introduced in this PR to common publishing schema would be added to `config`.

### DIFF
--- a/buildSrc/src/main/kotlin/fat-jar.gradle.kts
+++ b/buildSrc/src/main/kotlin/fat-jar.gradle.kts
@@ -25,14 +25,6 @@
  */
 
 import io.spine.internal.gradle.publish.SpinePublishing
-import org.gradle.api.Task
-import org.gradle.api.publish.maven.MavenPublication
-import org.gradle.kotlin.dsl.getValue
-import org.gradle.kotlin.dsl.getting
-import org.gradle.kotlin.dsl.java
-import org.gradle.kotlin.dsl.`maven-publish`
-import org.gradle.kotlin.dsl.project
-import org.gradle.kotlin.dsl.the
 
 plugins {
     `maven-publish`
@@ -44,10 +36,10 @@ plugins {
  * Creates a custom publication called `fatJar`, taking the output of `tasks.shadowJar`
  * as the sole publication artifact.
  *
- * This publication works in combination with exclusion of standard publishing for this
+ * This kind of publishing works in combination with exclusion of standard publishing for this
  * module specified in the root project under [SpinePublishing.modulesWithCustomPublishing].
  */
-project.publishing {
+publishing {
     publications {
         // The publishing settings from the root project.
         val spinePublishing = rootProject.the<SpinePublishing>()
--- a/buildSrc/src/main/kotlin/fat-jar.gradle.kts
+++ b/buildSrc/src/main/kotlin/fat-jar.gradle.kts
@@ -25,15 +25,19 @@
  */
 
 import io.spine.internal.gradle.publish.SpinePublishing
+import org.gradle.api.Task
+import org.gradle.api.publish.maven.MavenPublication
+import org.gradle.kotlin.dsl.getValue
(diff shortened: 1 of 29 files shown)


armiol:
@alexander-yevsyukov 
Something isn't right. The module in `master` is already renamed, as far as I can see. 

Also, if the name of the module is the same as the name of its fat JAR (and the fat JAR has no classifiers), their publishing will not be deterministic. In some cases, `pom.xml` of "small" artifact will overwrite the same file which belongs to "fat" artifact.

alexander-yevsyukov:
@armiol, PTAL. I how _now_ it should work. At least, publishing to `mavenLocal()` does without issues.

buildSrc/src/main/kotlin/io/spine/internal/gradle/RunGradle.kt
```diff
@@ -47,6 +47,15 @@ import org.gradle.internal.os.OperatingSystem
...
 
+    companion object {
+
+        /**
+         * Default Gradle build timeout.
+         */
+        private const val BUILD_TIMEOUT_MINUTES: Long = 10
+    }
+
+
```
armiol, line 58: This line is redundant.

armiol (approved):
LGTM. Not sure I understand how it works, but in any case the code became cleaner.

alexander-yevsyukov:
@dmdashenkov, PTAL too.

buildSrc/src/main/kotlin/fat-jar.gradle.kts
```diff
@@ -0,0 +1,99 @@
...
+         * Remove all third-party plugin declarations as well.
+         *
+         * They should be loaded from their respective dependencies.
+         */
+        "META-INF/gradle-plugins/com**",
+        "META-INF/gradle-plugins/net**",
+        "META-INF/gradle-plugins/org**")
+
+    isZip64 = true  /* The archive has way too many items. So using the Zip64 mode. */
+    archiveClassifier.set("")    /** To prevent Gradle setting something like `osx-x86_64`. */
```
armiol, line 88: If this is applied to all of our fat JARs, some of them (namely, artifacts defined in ProtoData and McJava) will probably lose their classifiers used currently.

Another thing is that we reference some of artifacts programmatically (McJava does that, at least, from what I recall). Therefore, any changes to coordinates, such as setting/removing the classifiers, should also be reflected in our code. 

That's something to consider.

armiol (commented):
@alexander-yevsyukov thanks for the documentation, but my understanding of Gradle publishing does not allow to understand why restricting the "preparations" to just assigning the Maven coordinates prevents the "small" artifact from publishing. So basically, I just know too little of this mechanism. And I can only believe it works by processing only the custom publication (fat JAR).

I have another comment inline for you to take into the account.

buildSrc/src/main/kotlin/fat-jar.gradle.kts
alexander-yevsyukov: @armiol, I suppose if those modules define their own `tasks.shadowJar`, they can "override" this behaviour when setting the classifier explicitly. They should be evaluated _after_ those tasks that are added by a script plugin. Don't they?

armiol: alexander-yevsyukov probably so, but with all these `afterSmth` tricks I would not be so sure. 

In any case, I would mention the classifier configuration in docs.

PR #84, opened 2023-04-10
Update ProtoData and include dependencies in the type system

### LINKED ISSUE #69: Kotlin compiler does not see generated proto DSL under `vanilla` module
We got the following error (in #68):

```
> Task :proto:configuration:generateTestRejections
e: /home/runner/work/validation/validation/java-tests/vanilla/src/test/kotlin/io/spine/validation/java/DistinctConstraintSpec.kt: (35, 26): Unresolved reference: constraintViolation

> Task :java-tests:vanilla:compileTestKotlin FAILED
e: /home/runner/work/validation/validation/java-tests/vanilla/src/test/kotlin/io/spine/validation/java/DistinctConstraintSpec.kt: (52, 24): Unresolved reference: constraintViolation
```

There is dependency on `java-runtime` that contains this declarations. And local builds under macOS pass normally.
We do want to have Kotlin DSL in tests. And this looks like an issue either in McJava or, which is more likely, in ProtoData.

dmdashenkov:
In this PR we make the validation's type system know about the module's dependencies. For that, we source info from the `ProtobufDependency` view.

We also make the runtime routine that unpacks `Any`s check if the type is known or now. If the type packed into the `Any` is unknown, the message is not validated and a warning is logged.

Also, fixes #69.


dmdashenkov:
@armiol, @alexander-yevsyukov, PTAL.

java-tests/vanilla/build.gradle.kts
```diff
@@ -65,3 +48,13 @@ dependencies {
     testImplementation(spine.base)
     testImplementation(spine.testlib)
 }
+
+sourceSets {
+    all {
+        java.setSrcDirs(java.filter { !it.absolutePath.startsWith("${buildDir.absolutePath}/generated") })
```
armiol, line 54: This piece will seem odd in a few months, or so. Could you please document the idea? I.e. why would we need to do that at all?

armiol (changes requested):
@dmdashenkov LGTM except for a single comment. That one may lead to others, once addressed.

alexander-yevsyukov (approved):
Have a look at a couple of FYI notes I added. I concur with the request from @armiol on documenting the filtering code. Apart from that, LGTM. Looking forward to the further, much wanted, improvements.

dmdashenkov:
@armiol, PTAL.

armiol (approved):

PR #85, opened 2023-06-07
Update config

alexander-yevsyukov:


alexander-yevsyukov:
@dmdashenkov, just in case you'd want to update dependencies for Validation, here's the PR which I already created for that.

The problem is that it does not pass integration tests because of dependency issue. I'm going to re-build it, once GitHub Pull Requests outage is fixed, and newer `model-compiler` is published. Probably, it would require updating `mc-java` too AND rebuilding Validation once again.

This is just FYI.

alexander-yevsyukov:
Closing as outdated.

PR #88, opened 2023-06-21
Explicit task dependencies

dmdashenkov:
In this PR we add explicit dependencies between the `jar` task and the `publishFatJarTo<Repo>Repository`.

This change effects for `-bundle` subprojects.

We do this in an effort to fix the publication scripts which turned out to be broken, as the `-bundle` artifacts were not published in the latest PR.


dmdashenkov:
@armiol, PTAL.

armiol (approved):
LGTM with a suggestion.

PR #89, opened 2023-06-22
Fix artifact ID for fat jars

dmdashenkov:
In this PR we fix the custom artifact configuration so that the artifact ID is only set once.

Previously, bundle artifacts used to be published with double prefix: `spine-validation-spine-validation-java-bundle` instead of `spine-validation-java-bundle`.


dmdashenkov:
@alexander-yevsyukov, @armiol, PTAL.

alexander-yevsyukov (approved):
LGTM

PR #92, opened 2023-07-07
Improve insertion point performance

dmdashenkov:


buildSrc/src/main/kotlin/module.gradle.kts
```diff
@@ -24,11 +24,13 @@
  * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
+import Module_gradle.Module
```
alexander-yevsyukov, line 27: @dmdashenkov, do we really need this import?

PR #93, opened 2023-07-11
Performance improvements

dmdashenkov:
In this PR we make the validation's insertion points repeated. This reduces the number of operations required to lookup insertion points' desired positions and improves performance.

For further build time speed up, we add an extra level of caching for the Roaster's model to avoid unnecessary code parsing.

Also, `config` is updated.


java/src/main/kotlin/io/spine/validation/java/CachingJavaClassSource.kt
```diff
@@ -0,0 +1,53 @@
...
+ * A `JavaClassSource` which caches some of its parsed components.
+ *
+ * Please note that, because of the caching, sources updated in an underlying file may not be
+ * updated in this model.
+ */
+internal class CachingJavaClassSource(
+    private val delegate: JavaClassSource
+) : JavaClassSource by delegate {
+
+
```
alexander-yevsyukov, line 42: @dmdashenkov, please remove this empty line.

dmdashenkov:
@armiol, you may now take this branch and run your performance tests with it. On my local machine, the `BuildSpeed` time is now ~2:02 minutes, compared to ~4 minutes before the insertion point optimizations.

dmdashenkov:
@armiol, PTAL. The build is failing at the moment because of a JaCoCo issue, unrelated to the core idea of the PR. Will be resolved later (before the merge).

java/src/main/proto/spine/validation/java/type.proto
```diff
@@ -0,0 +1,48 @@
...
+option (type_url_prefix) = "type.spine.io";
+option java_package = "io.spine.validation.test";
+option java_outer_classname = "BlizzardProto";
+option java_multiple_files = true;
+option (internal_all) = true;
+
+import "spine/protodata/ast.proto";
+
+// A tuple of a message type and the file where it is declared.
+//
```
alexander-yevsyukov, line 42: We agreed not to have a trailing empty line for one-line proto documentation. Please remove.

java/src/main/proto/spine/validation/java/type.proto
```diff
@@ -0,0 +1,48 @@
...
+message MessageWithFile {
+
+    spine.protodata.MessageType message = 1;
```
alexander-yevsyukov, line 45: @armiol, @dmdashenkov, similarly to Validation, we intended to market ProtoData as a relatively independent product.

What if we have proto packages starting from `protodata` (instead of `spine.protodata`)?

alexander-yevsyukov: This question does not relate to this PR, but I take the chance to discuss it here.

armiol: @dmdashenkov @alexander-yevsyukov let's discuss it in a meeting.

My vote is for `spine.` removal here. And! Somewhen we'll need to come up with a name for the Validation library, so that we could drop the `spine.` prefix there as well.

armiol (approved):
@dmdashenkov LGTM from my side.

Please have a look at the comments by @alexander-yevsyukov, though.

java/src/main/proto/spine/validation/java/type.proto
alexander-yevsyukov: Our validation options which currently belong to `base`. They do not have a package for brevity of declarations at usage sites.  So, we do not have the longer prefix issue for these types at all.

Someday `options.proto` may become a part of the Validation library (thus making `base` depend on `validation`) for the sake of cohesion of the types. But it's not a priority now. I'm talking about this in the context of discussing types of the Validation library.

alexander-yevsyukov (approved):
LGTM

PR #100, opened 2023-09-04
Update ProtoData to `0.11.4`

alexander-yevsyukov:
This PR updates the dependency on ProtoData, addressing appropriate deprecations.

**Other notable changes:**
 * The order of parameters in `GenerationContext` was improved.


armiol (approved):
@alexander-yevsyukov LGTM 💯.
Also, next time we should aim for bumping the dogfooding version as well. But let's not spoil this PR#100 :)

PR #107, opened 2023-11-09
Start adopting Validation codegen

alexander-yevsyukov:
This PR is the first of the seria brining code-generation of validation to life.

Here's the sequence of changes that we made in relation to this:

1. The validation codegen was turned on via removing `skipValidation()` calls in build files. 
2. Several issues in codegen were discovered.
3. The issues were fixed and tested via local deployment of Validation library and subsequent update and local deployment of McJava.
5. Then, the `skipValidation()` was turned back on, and McJava version rolled back to `2.0.0-SNAPSHOT.170`, so that we can build it at GitHub. 
6. This, of course, resulted in usage of older code of Validation, which required to disable tests that were written for newer validation code.

### Further actions
 * Publish artifacts of this PR.
 * Assemble new version of McJava which uses this version of Validation.
 * Extend McJava allowing to use custom version of Validation, so that we can use McJava with the locally deployed version being built.
 * Turn test annotated with `@Disabled("Until 'skipValidation()` is turned off.")` back on.

### Other notable changes
 * Support of external validation constraints was removed.
 * Tests that required updates because of changes in codegen, were migrated to Kotlin.
 * A temporary `Workaround.kt` was introduced to allow using run-time validation for features that are not yet covered by codegen.
 * `Diags` class was introduced for gathering diagnostic messages and routines in one place. This is needed for easier introduction of localization in the future.
 * Kotlin proto DSL was adopted in the modified Kotlin code.
 * `codeBlock { }` DSL was adopted.
 * Stub proto types were separated by a file per test suite.
 * Internal and external dependencies were updated.


java-runtime/src/test/kotlin/io/spine/validate/EnclosedMessageValidationSpec.kt
```diff
@@ -0,0 +1,149 @@
...
+        val innerViolation = innerViolations[0]
+
+        innerViolation.msgFormat shouldStartWith Diags.Regex.prefix
+
+        assertFieldPathIs(
+            innerViolation,
+            EMAIL
+        )
+
+        innerViolation.violationList.shouldBeEmpty()
```
yevhenii-nadtochii, line 121: There are many empty lines, and it is quite hard to understand grouping. Maybe reduce number of empty lines or extract smaller asserting methods?

java-runtime/src/test/java/io/spine/validate/option/RequiredTest.kt
```diff
@@ -0,0 +1,160 @@
...
+import io.spine.validate.ValidationOfConstraintTest.Companion.VALIDATION_SHOULD
+import io.spine.validate.given.MessageValidatorTestEnv
+import io.spine.validate.given.MessageValidatorTestEnv.newByteString
+import io.spine.validate.given.MessageValidatorTestEnv.newStringValue
+import org.junit.jupiter.api.Disabled
+import org.junit.jupiter.api.DisplayName
+import org.junit.jupiter.api.Test
+
+@DisplayName(VALIDATION_SHOULD + "analyze `(required)` option and")
+internal class RequiredTest : ValidationOfConstraintTest() {
```
yevhenii-nadtochii, line 52: Should it have been called `RequiredSpec`?

java-runtime/src/test/java/io/spine/validate/option/RequiredTest.kt
```diff
@@ -0,0 +1,160 @@
+/*
+ * Copyright 2022, TeamDev. All rights reserved.
```
yevhenii-nadtochii, line 2:
```suggestion
 * Copyright 2023, TeamDev. All rights reserved.
```

yevhenii-nadtochii: Though, most files have `2023` already.

java-runtime/src/test/java/io/spine/validate/option/RequiredTest.kt
```diff
@@ -0,0 +1,160 @@
...
+ * OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+package io.spine.validate.option
+
```
yevhenii-nadtochii, line 27: We should separate `package` declaration from a copyright notice.

java-runtime/src/test/kotlin/io/spine/validate/NumberRangeSpec.kt
```diff
@@ -0,0 +1,182 @@
...
+    fun `find out that number is equal to max inclusive`() =
+        maxNumberTest(EQUAL_MAX, inclusive = true, valid = true)
+
+    @Test
+    fun `find out that number is less than max inclusive`() =
+        maxNumberTest(LESS_THAN_MAX, inclusive = true, valid = true)
+
+    @Test
+    fun `find out that number is less than max non-inclusive`() =
+        maxNumberTest(GREATER_THAN_MAX, inclusive = false, valid = false)
```
yevhenii-nadtochii, line 97: Can it be more straightforward?

```kotlin
maxNumberTest(LESS_THAN_MAX, inclusive = false, valid = true)
```

java-runtime/src/test/kotlin/io/spine/validate/option/GoesSpec.kt
```diff
@@ -0,0 +1,104 @@
...
+
+        @Test
+        fun `are both optional` () = assertValid {
+            payment {
+                description = "Scheduled payment"
+            }
+        }
+
+        @Test
+        fun `not filled simultaneously` () {
```
yevhenii-nadtochii, line 59:
```suggestion
        fun `not filled separately ` () {
```

java-runtime/src/test/java/io/spine/validate/option/RequiredTest.kt
yevhenii-nadtochii: And there are many such files down.

yevhenii-nadtochii (commented):
Please see my comments.

Although, I haven't viewed all files yet.

yevhenii-nadtochii (approved):
LGTM in general.

I've left minor suggestions to consider before merging.

java/src/main/kotlin/io/spine/validation/java/ValidationConstraintsCode.kt
```diff
@@ -0,0 +1,134 @@
...
+
+    /**
+     * Obtains the generated code block.
+     */
+    fun codeBlock(): CodeBlock = code.build()
+
+    /**
+     * Obtains class-level declarations used in the validation code as code lines.
+     */
+    fun supportingMembersCode(): ImmutableList<String> {
```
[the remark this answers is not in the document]
alexander-yevsyukov: It's not really. The method is used only once in `ValidationCode`, so I'm going to change it to `Iterable` instead.

model/src/main/kotlin/io/spine/validation/NumberRules.kt
```diff
@@ -97,34 +106,28 @@ private constructor(
...
+            fun Value.str(): String = when (kindCase) {
+                DOUBLE_VALUE -> doubleValue.toString()
+                INT_VALUE -> intValue.toString()
+                else -> error("Unexpected Value: `$this`.")
+            }
+
+            return "The number must be between ${lowerBound!!.str()} " +
+                    "(${lowerInclusive.str()}) and ${upperBound!!.str()} " +
+                    "(${uppedInclusive.str()}), but was {value}."
+        }
```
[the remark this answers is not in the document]
alexander-yevsyukov: It is a property in a sense that its value does not change. It's hardly wired to other `val` properties of this instance.

I admit it's not a trivial piece of code, but it doesn't change the nature of what's going on there. 

I'll make it `by lazy` to highlight the "propertness".

java-runtime/src/test/kotlin/io/spine/validate/NumberRangeSpec.kt
alexander-yevsyukov: The first parameter is a value we pass to tests. The `valid` parameter when `true` tells that failure should be asserted.

Here we pass invalid value, and want an error to be raised.

PR #109, opened 2023-12-08
Migrate to updated ProtoData API

alexander-yevsyukov:
This PR migrates Validation to ProtoData v0.15.0, which introduces breaking API improvements (as compared with `0.14.3`).

### DIFF
--- a/buildSrc/src/main/kotlin/ForcePlugins.kt
+++ b/buildSrc/src/main/kotlin/ForcePlugins.kt
@@ -47,6 +47,9 @@ fun ScriptHandlerScope.forceCodegenPlugins() {
     configurations.forceProtoData()
 }
 
+/**
+ * Forces the version of McJava and ProtoData for a module with this `ConfigurationContainer`.
+ */
 fun ConfigurationContainer.forceProtoData() = all {
     resolutionStrategy {
         val protoData = io.spine.internal.dependency.ProtoData
--- a/buildSrc/src/main/kotlin/ForcePlugins.kt
+++ b/buildSrc/src/main/kotlin/ForcePlugins.kt
@@ -24,14 +24,37 @@
  * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
-buildscript {
-    forceCodegenPlugins()
+import org.gradle.api.artifacts.ConfigurationContainer
+import org.gradle.kotlin.dsl.ScriptHandlerScope
+
+/**
+ * Forces the version of McJava and ProtoData for a module.
+ *
+ * The usage scenario is put the following code snippet at the top of a `build.gradle.kts` file:
+ *
+ * ```kotlin
+ * buildscript {
+ *     forceCodegenPlugins()
+ * }
+ * ```
+ */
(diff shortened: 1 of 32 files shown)


buildSrc/src/main/kotlin/ForcePlugins.kt
```diff
@@ -0,0 +1,60 @@
...
+fun ScriptHandlerScope.forceCodegenPlugins() {
+    standardSpineSdkRepositories()
+    dependencies {
+        classpath(io.spine.internal.dependency.Spine.McJava.pluginLib)
+        classpath(io.spine.internal.dependency.ProtoData.pluginLib)
+    }
+    configurations.forceProtoData()
+}
+
+fun ConfigurationContainer.forceProtoData() = all {
```
armiol, line 50: Please document this extension as well.

java-tests/extensions/build.gradle.kts
```diff
@@ -24,6 +24,18 @@
...
+}
+
 dependencies {
     implementation(project(":java"))
 }
+
+modelCompiler {
+    java {
+        codegen {
+            rejections().enabled.set(false)
```
armiol, line 38: Let's leave a note to our descendants here :)

E.g.
```
The generation of rejections is disabled, 
since we don't want other plugins (potentially, 
still not using the latest ProtoData API) 
interfere with the tests of Validation ProtoData plugin.
```

java-tests/extra-definitions/build.gradle.kts
armiol, line 1: This `extra-definitions` module seems new. I am not sure what it does — and the PR description is also silent on this matter.

Let's discuss first.

java/src/main/proto/spine/validation/java/type.proto
```diff
@@ -43,5 +43,5 @@ message MessageWithFile {
 
     spine.protodata.MessageType message = 1;
 
-    spine.protodata.File declared_in = 2;
+    spine.protodata.ProtoFileHeader declared_in = 2;
```
armiol, line 46: Let's rename this field, as types are not "declared" in headers.

model/src/main/java/io/spine/validation/DistinctPolicy.java
```diff
@@ -41,7 +41,7 @@
 
 import static io.spine.protobuf.AnyPacker.unpack;
 import static io.spine.protodata.Ast.isRepeated;
-import static io.spine.protodata.Ast.qualifiedName;
+import static io.spine.protodata.Ast.getQualifiedName;
```
armiol, line 44: This also looks weird.

I realize it's now too late to comment on this, but I definitely liked the version without the getter better.

armiol (changes requested):
@alexander-yevsyukov please see my comments.

model/src/main/java/io/spine/validation/DistinctPolicy.java
alexander-yevsyukov: `qualifiedName` used to be a method. Now it's a property, which looks more native in Kotlin code. Since we're in Java in this class, we use the automatic getter for the property.

armiol (approved):
LGTM.

PR #110, opened 2023-12-26
Adopt changes to `TypeName` API from ProtoData 

alexander-yevsyukov:
This PR updates the dependency on ProtoData thus fixing the compatibility issue on the extension function from `Expressions` class of ProtoData. 

Previously, the extension was `public fun ClassName.call()`, while latest ProtoData has it as `public fun ClassOrEnumName.call()`. Even though, `ClassOrEnumName` is a subtype of `ClassName`, the Java implementation of the extensions is not backward compatible. 

Also This PR updates versions of `tool-base`, thus brining PSI into the dependencies.


armiol (approved):
@alexander-yevsyukov LGTM.

PR #127, opened 2024-07-12
Implement `ValidatingBuilder` instead of McJava

alexander-yevsyukov:
This PR makes the Validation library make `Builder` classes implement `ValidatingBuilder` instead of McJava.

Currently, McJava does that via the `io.spine.tools.mc.java.protoc.message.BuilderGen` and associated classes which work in the `protoc` plugin of McJava. 

This PR introduces `ImplementValidatingBuilder` renderer which makes `Builder` classes implement `ValidatingBuilder`. Adding the code is done only if the `Builder` class does not implement the interface already. Such a case is possible when older McJava is used together with new Validation. Once McJava migrates to newer Validation, the check would be always `false`.

### DIFF
--- a/java/src/main/kotlin/io/spine/validation/java/JavaRendererExts.kt
+++ b/java/src/main/kotlin/io/spine/validation/java/JavaRendererExts.kt
@@ -1,0 +1,52 @@
+/*
+ * Copyright 2024, TeamDev. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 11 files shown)


java/src/main/kotlin/io/spine/validation/java/JavaRendererExts.kt
```diff
@@ -0,0 +1,52 @@
...
+ * Obtains a collection of messages from the given source file paired with the file header.
+ */
+private fun ProtobufSourceFile.messages(): Collection<MessageWithFile> =
+    typeMap.values.map {
+        messageWithFile {
+            message = it
+            fileHeader = header
+        }
+    }
+
```
armiol, line 52: This line is redundant.

armiol (approved):
@alexander-yevsyukov LGTM with a minor comment.

PR #128, opened 2024-07-16
Bump McJava and Validation

alexander-yevsyukov:
This PR migrates to newer McJava, which no longer makes `Message.Builder` implement `ValidatingBuilder` interface.

Because of this change, codegen for the backend now have to use Validation features, and its version also bumped.

### DIFF
--- a/proto/context/build.gradle.kts
+++ b/proto/context/build.gradle.kts
@@ -5,7 +5,7 @@
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
  *
- * http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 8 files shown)


proto/context/build.gradle.kts
```diff
@@ -59,10 +59,10 @@ dependencies {
...
 // and the generated code.
 //
-modelCompiler {
-    java {
-        codegen {
-            validation().enabled.set(false)
-        }
-    }
-}
+//modelCompiler {
```
armiol, line 62: If the whole thing works together with tests, maybe we could kill these commented portions in this and other files?

armiol (approved):
@alexander-yevsyukov LGTM, with a comment to consider.

PR #129, opened 2024-07-20
Filter out gRPC generated code when adding `implements ValidatingBuilder`

alexander-yevsyukov:
This PR fixes filtering source sets that are output of gRPC compiler. `ImplementValidatingBuilder` action does not apply to these sources and fail with an exception when a Java file could not be found.

Previously, the code of `ImplementValidatingBuilder` action relied on the call to `SourceFileSet.hasJavaFiles`, which checks for the presence of a Java file in the set. gRPC Protobuf compiler does produce Java files, but the output placed under the `grpc` directory does not contain message files participating in gRPC calls. They are placed under `java` directory. Because of this it is not possible to find a source code file matching a message in the source file set with the `grpc` output directory.

Now, we call `SourceFileSet.hasJavaOutput`, which checks the target directory and that solves the filtering issue.

### DIFF
--- a/java/build.gradle.kts
+++ b/java/build.gradle.kts
@@ -28,7 +28,6 @@ import io.spine.internal.dependency.JavaPoet
 import io.spine.internal.dependency.Kotest
 import io.spine.internal.dependency.ProtoData
 import io.spine.internal.dependency.Roaster
-import io.spine.internal.dependency.Spine
 
 plugins {
     `build-proto-model`
--- a/java/build.gradle.kts
+++ b/java/build.gradle.kts
@@ -5,7 +5,7 @@
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
  *
- * http://www.apache.org/licenses/LICENSE-2.0
+ * https://www.apache.org/licenses/LICENSE-2.0
  *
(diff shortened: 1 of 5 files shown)


java/build.gradle.kts
```diff
@@ -28,6 +28,7 @@ import io.spine.internal.dependency.JavaPoet
 import io.spine.internal.dependency.Kotest
 import io.spine.internal.dependency.ProtoData
 import io.spine.internal.dependency.Roaster
+import io.spine.internal.dependency.Spine
```
armiol, line 31: This is weird. Nothing has changed in the script itself. but the import. Please review.

armiol (approved):
@alexander-yevsyukov LGTM, but please see my comment.

Also, I don't understand how this PR addresses filtering out the gRPC-generated sources. I am sure it works :) But let's discuss how :)

alexander-yevsyukov:
> @alexander-yevsyukov LGTM, but please see my comment.
> 
> Also, I don't understand how this PR addresses filtering out the gRPC-generated sources. I am sure it works :) But let's discuss how :)

It's indeed not obvious. I've. updated the description of the PR.

PR #137, opened 2024-10-10
Migrate to new `Plugin` API

alexander-yevsyukov:
This PR migrates Validation plugins to new `Plugin` API recently introduced in ProtoData.

Convenience extension functions `asA()` and `asB()` were also adopted.

### DIFF
--- a/model/src/main/java/io/spine/validation/ValidationPolicy.java
+++ b/model/src/main/java/io/spine/validation/ValidationPolicy.java
@@ -49,23 +49,4 @@ public abstract class ValidationPolicy<E extends EventMessage>
     @Override
     @ContractFor(handler = React.class)
     protected abstract EitherOf2<RuleAdded, NoReaction> whenever(E event);
-
-    /**
-     * Creates an {@link EitherOf2} with {@link NoReaction} in the {@code B} option.
-     *
-     * <p>Usage example:
-     * <pre>
-     *  {@literal class MyPolicy extends ValidationPolicy<TypeEntered>} {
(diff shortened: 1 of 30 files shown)


model/src/main/java/io/spine/validation/ValidationPolicy.java
```diff
@@ -48,25 +48,24 @@ public abstract class ValidationPolicy<E extends EventMessage>
...
+     *               return ignoring();
      *           }
      *           return myCustomRule(event);
      *       }
      * </pre>
      */
-    protected final EitherOf2<RuleAdded, Nothing> noReaction() {
-        //TODO:2024-08-11:alexander.yevsyukov: Use EventProducer.noReaction() extension from `core-java`.
-        return EitherOf2.withB(nothing());
+    protected final EitherOf2<RuleAdded, NoReaction> ignoring() {
```
armiol, line 68: See my comments above. And let's discuss.

armiol (changes requested):
@alexander-yevsyukov please see my comments, and let's discuss.

yevhenii-nadtochii (approved):
LGTM from my side.

I would also go with `noReaction()` because it matches the method signature of the reactor.

armiol (approved):
@alexander-yevsyukov LGTM.

@yevhenii-nadtochii FYI. We have settled on `ignore()`, as `noReaction()` is impossible to have to return `Either...` instance — in the same class hierarchy it is already busy returning plain `NoReaction`. So there's that.

PR #138, opened 2024-10-21
Support `(set_once)` in validation codegen

yevhenii-nadtochii:
This PR extends coverage of validation codegen to support  `(set_once)` constraint.

The option is enforced by checking (during a field value assignment) that the current value is either default for the field type OR it is the same with the assigned one. Otherwise, the validation exception is thrown **right away**. 

When one or more fields with the message are marked with the option, the following groups of methods are modified to add the precondition checks:

1. All `set...()` methods for the marked field.
2. All `merge...()` methods, except for `AbstractMessage.mergeFrom(final Message other)`. In merge methods, only parts relevant to the marked fields are affected.

### `AbstractMessage.mergeFrom(final Message other)`

This method is not **explicitly supported**, though it might be workable. 

### Supported field types

All primitives, enums and messages are supported. 

An attempt to mark `repeated` or `map` field leads to an exception.


yevhenii-nadtochii:
@armiol @alexander-yevsyukov 

We should settle down on the option specification before proceeding with this PR. I've examined the generated code, and the main disappointment is that Protobuf v3 doesn't store explicit information about whether the field was set (except for `optional` fields). Our implementation checks that the current field value is default for the field type, only then allowing a new one to set.

In general, Protobuf builders provide three paths to modify the message.

`set...()`:

1. Direct field setter like `setName(String name)`.
2. Alternative setters like `setName(ByteString name)` for `string` and `setMyEnumValue(int value)` for enums.
2. Direct field setter that accepts a builder instead of a message (only for message-based fields) like `setStudent(StudentBuilder builder)`.
6. Reflective setter, which sets a field by its descriptor `setField(FieldDescriptor, Object)`.

`merge...()`:

1. Field merging (only for message-based fields) like `mergeName(Name value)`.
2. Message merging `mergeFrom(Student other)`.
3. Bytes merging `mergeFrom(byte[] data)`.

`clear...()`:

1. Clear a specific field `clearName()`.
2. Clear the whole message `clear()`.

The questions are the following:

1. Should we allow a custom error message for this option? 
3. Which field types are supported? - I suggest messages, enumerations and all primitives excluding repeated fields and maps. `bytes` arrays are also allowed.
4. Which points of message mutation does the option cover? – All setters for sure. But should we cover merging and clearing? The current implementation implicitly covers because it just compares a field value having two message instances. The problem is that for many field types, the merging is based on setters. We have to overcome this. And clearing allows overriding a field value, for which `(set_once)` is specified.
5. Should we allow the field to be set to the current value? – Looks logical if we cover `merge`-based mutations. Looks safe for setters as well.
6. Should we prohibit clearing of a specific field if it is marked with the option? What about the builder `clear()` then? – I suggest considering `merge` and `clear` service methods, and don't touch them at all. Only fix the fact that for some types, `merge` redirect to setters. We can introduce private unsafe setters for that, for example.

alexander-yevsyukov:
> 1. Should we allow a custom error message for this option?

This is highly desirable. People creating their domain language would want to explain why a field can be set only once. These cases are usually non-trivial.

alexander-yevsyukov:
> 2. Which field types are supported? - I suggest messages, enumerations and all primitives excluding repeated fields and maps. bytes arrays are also allowed.

I concur.

alexander-yevsyukov:
> 3. Which points of message mutation does the option cover? – All setters for sure. But should we cover merging and clearing? 

Merging — it is desirable.

See on clearing below.

> ... And clearing allows overriding a field value, for which (set_once) is specified.

The `(set_once)` option largely protects from setting the value via a setter. If the developer explicitly calls `clear()` it corresponds to starting all over on this message. Something happened in the domain, which cased the code to clear the message. To me it close to having `newBuilder()`. I suggest viewing `clear()` this way.

alexander-yevsyukov:
> 4. Should we allow the field to be set to the current value? 

I think this way it's understandable and safe. We should allow this, provided this is documented in the option documentation.

alexander-yevsyukov:
> 5. Should we prohibit clearing of a specific field if it is marked with the option? 

Let's allow clearing such fields. Please also see my earlier comment regarding the clear operation.

> What about the builder clear() then? 

We should allow this. See my earlier comment on the "weight" of the clear operation.

@armiol, FYI.

yevhenii-nadtochii:

java-runtime/src/main/java/io/spine/validate/Validate.java
```diff
@@ -208,7 +210,9 @@ public static <M extends Message> void checkValidChange(M previous, M current) {
      *         the type of the message
      * @return a set of constraint violations, if the transaction is invalid,
      *         an empty set otherwise
+     * @deprecated the {@code set_once} constraint is enforced within the message builder now.
```
alexander-yevsyukov, line 213: Again, as per my comment above, we don't give the user instructions on what he should do about the deprecation. Please rephrase.

java-tests/validation-gen/src/test/java/io/spine/test/tools/validate/setonce/package-info.java
```diff
@@ -0,0 +1,27 @@
...
+ * OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+package io.spine.test.tools.validate.setonce;
```
alexander-yevsyukov, line 27: Please provide package annotations we have for the Java packages in the production code.

Also a package description would be nice.

java/src/main/kotlin/io/spine/validation/java/setonce/PsiClassExts.kt
```diff
@@ -0,0 +1,53 @@
...
+/**
+ * Returns a method from this [PsiClass] matching the given signature
+ * specified as [text].
+ *
+ * @param text The method signature as text.
+ * @throws IllegalStateException if this class does not have such a method.
+ */
+internal fun PsiClass.getMethodBySignature(text: String): PsiMethod =
+    findMethodBySignature(text)
+        ?: error("Could not find method `$text` in `$qualifiedName` class.")
```
alexander-yevsyukov, line 53: "Could not find the method with the signature `$text` in the `$qualifiedName` class."

java/src/main/kotlin/io/spine/validation/java/setonce/PsiElementFactoryExts.kt
```diff
@@ -0,0 +1,43 @@
...
+
+import com.intellij.psi.PsiElementFactory
+import com.intellij.psi.PsiMethod
+import com.intellij.psi.PsiStatement
+
+/**
+ * Creates a new [PsiMethod] from the given [text].
+ */
+internal fun PsiElementFactory.createMethod(text: String): PsiMethod =
+    createMethodFromText(text, null)
```
alexander-yevsyukov, line 37: I understand the desire for brevity on the code, but I think we should keep using the context of the method creation as per standard PSI API.

alexander-yevsyukov: We now don't use the context in full because we use trimmed down version of the project. But we want to have more featured project in the future. One of the applications of this would be optimising imports in the code.

So, not using the context now will require returning them later.

java/src/main/kotlin/io/spine/validation/java/setonce/SetOnceJavaConstraints.kt
```diff
@@ -0,0 +1,214 @@
...
+
+/**
+ * Renders Java code to support `(set_once)` option for the given [field].
+ *
+ * The rendered Java constraints are specific to the field type. This class
+ * serves as an abstract base, providing common methods and the skeleton implementation
+ * of [render] method. Inheritors should perform actual rendering in [renderConstraints].
+ *
+ * @property field The field that declared the option.
+ * @property messageWithHeader The message that contains the [field].
```
alexander-yevsyukov, line 53: Please rename to `declaredIn`. Assuming the usage of this property it would be shorter and look a bit more readable.

java/src/main/kotlin/io/spine/validation/java/setonce/SetOnceJavaConstraints.kt
```diff
@@ -0,0 +1,214 @@
...
+     *
+     * Implementation of this method is common for all field types. Inheritors should
+     * invoke it within [renderConstraints], passing the necessary parameters.
+     *
+     * The [currentValue] and [readerStartsWith] are mandatory properties. Pass [readerContains]
+     * in cases when [readerStartsWith] is not sufficient. For example, for message fields,
+     * the beginning of the reading block is the same for all fields because it doesn't include
+     * the field name. Differentiation is done way deeper.
+     *
+     * @param currentValue an expression to read the current field value.
```
alexander-yevsyukov, line 141: Please capitalise parameter descriptions.

java/src/main/kotlin/io/spine/validation/java/setonce/SetOnceJavaConstraints.kt
```diff
@@ -0,0 +1,214 @@
...
+        fieldProcessing.addAfter(postcondition, fieldReading)
+    }
+
+    /**
+     * Creates an `if` statement, which checks that the current field value is default
+     * OR if the proposed value is the same as the current.
+     *
+     * Otherwise, it throws the validation exception.
+     *
+     * @param currentValue an expression denoting the current field value.
```
alexander-yevsyukov, line 171: Please capitalise parameter descriptions.

alexander-yevsyukov (changes requested):
Please see my comments.

java-tests/validation-gen/src/test/java/io/spine/test/tools/validate/setonce/SetOnceFieldsTest.java
```diff
@@ -0,0 +1,811 @@
...
+        }
+    }
+
+    @Nested
+    @DisplayName("allow overriding default and same-value double")
+    class AllowOverridingDefaultAndSameValueDouble {
+
+        private static final double halfOfMeter = 0.5;
+
+        private final Student unheightedStudent = Student.newBuilder()
```
armiol, line 348: `unmeasuredStudent`?

java-tests/validation-gen/src/test/java/io/spine/test/tools/validate/setonce/SetOnceFieldsTest.java
```diff
@@ -0,0 +1,811 @@
...
+    }
+
+    @Nested
+    @DisplayName("prohibit overriding non-default float")
+    class ProhibitOverridingNonDefaultFloat {
+
+        private static final float fiftyKilograms = 0.5f;
+
+        private final Student heavyStudent = Student.newBuilder()
+                .setWeight(80)
```
armiol, line 407: If `0.5f` is "fifty kilograms", then "80" becomes "eight tons"?

java-tests/validation-gen/src/test/java/io/spine/test/tools/validate/setonce/SetOnceFieldsTest.java
```diff
@@ -0,0 +1,811 @@
...
+                    .setHeight(halfOfMeter)
+                    .build());
+        }
+    }
+
+    @Nested
+    @DisplayName("prohibit overriding non-default float")
+    class ProhibitOverridingNonDefaultFloat {
+
+        private static final float fiftyKilograms = 0.5f;
```
armiol, line 404: `fiftyKg`?

java-tests/validation-gen/src/test/java/io/spine/test/tools/validate/setonce/SetOnceFieldsTest.java
```diff
@@ -0,0 +1,811 @@
...
+            assertValidationFails(() -> heavyStudent.toBuilder()
+                    .mergeFrom(thinStudent.toByteArray()));
+        }
+    }
+
+    @Nested
+    @DisplayName("allow overriding default and same-value float")
+    class AllowOverridingDefaultAndSameValueFloat {
+
+        private static final float fiftyKilograms = 0.5f;
```
armiol, line 446: I would have these repeating values moved to a corresponding test environment.

It would simplify making any changes, such as renaming, or changing the value.

java-tests/validation-gen/src/test/java/io/spine/test/tools/validate/setonce/SetOnceFieldsTest.java
```diff
@@ -0,0 +1,811 @@
...
+                    .setHeight(fiftyKilograms)
+                    .build());
+        }
+    }
+
+    /**
+     * Please note, for `boolean` fields, there are no `byMessageMerge` and `byBytesMerge` tests.
+     *
+     * <p>It is impossible to override a non-default value by another non-default value
+     * for a `boolean` field. In Protobuf v3, `boolean` has only one non-default value: `true`.
```
armiol, line 504: In Java, we don't use back-ticks. We use `{@code}`.

java-tests/validation-gen/src/test/java/io/spine/test/tools/validate/setonce/SetOnceFieldsTest.java
```diff
@@ -0,0 +1,811 @@
...
+        void afterClearing() {
+            assertValidationPasses(() -> thinStudent.toBuilder()
+                    .clearHeight()
+                    .setHeight(fiftyKilograms)
+                    .build());
+        }
+    }
+
+    /**
+     * Please note, for `boolean` fields, there are no `byMessageMerge` and `byBytesMerge` tests.
```
armiol, line 501: Please add some context, as a paragraph before this one.

java-tests/validation-gen/src/test/java/io/spine/test/tools/validate/setonce/SetOnceFieldsTest.java
```diff
@@ -0,0 +1,811 @@
...
+     *
+     * <p>It is impossible to override a non-default value by another non-default value
+     * for a `boolean` field. In Protobuf v3, `boolean` has only one non-default value: `true`.
+     * When we try to override `true` with `false` by merging, the merge method does nothing because
+     * it doesn't consider fields with the default values. When we override `false` with `true`,
+     * we're just effectively assigning an initial non-default value, which is tested by another
+     * test suite. See {@link AllowOverridingDefaultAndSameValueBoolean}.
```
armiol, line 508: Maybe, use the standard `@see` tag instead?

java-tests/validation-gen/src/test/java/io/spine/test/tools/validate/setonce/SetOnceFieldsTest.java
```diff
@@ -0,0 +1,811 @@
...
+                .build();
+        private final Student awardedStudent = Student.newBuilder()
+                .setHasMedals(true)
+                .build();
+
+        @Test
+        @DisplayName("by value")
+        void byValue() {
+            assertValidationPasses(() -> studentWithoutMedals.toBuilder()
+                    .setHasMedals(has)
```
armiol, line 551: I would say `.setHasMedals(yes)`.

java-tests/validation-gen/src/test/java/io/spine/test/tools/validate/setonce/SetOnceIntegerFieldsTest.java
```diff
@@ -0,0 +1,1042 @@
...
+import static io.spine.test.tools.validate.setonce.SetOnceTestEnv.assertValidationPasses;
+
+/**
+ * Tests {@code (set_once)} constraint for integer fields.
+ */
+@DisplayName("`(set_once)` constraint should")
+class SetOnceIntegerFieldsTest {
+
+    @Nested
+    @DisplayName("prohibit overriding non-default int32")
```
armiol, line 44: Let's improve the references to the type a bit:
```java
@DisplayName("prohibit overriding non-default `int32`")
```

Same applies to the rest of the cases.

java-tests/validation-gen/src/test/java/io/spine/test/tools/validate/setonce/SetOnceTestEnv.java
```diff
@@ -0,0 +1,60 @@
...
+        assertThrows(ValidationException.class, executable);
+    }
+
+    /**
+     * Asserts that the given {@code executable} doesn't throw anything.
+     */
+    static void assertValidationPasses(Executable executable) {
+        assertDoesNotThrow(executable);
+    }
+
```
armiol, line 59: Let's kill this line.

java-tests/validation-gen/src/test/java/io/spine/test/tools/validate/setonce/SetOnceTestEnv.java
```diff
@@ -0,0 +1,60 @@
...
+import io.spine.validate.ValidationException;
+import org.junit.jupiter.api.function.Executable;
+
+import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+/**
+ * Test environment for {@code (set_once)} tests.
+ */
+class SetOnceTestEnv {
```
armiol, line 38: The methods in this class are hardly a test-environment grade. They aren't designed to describe the corresponding test environment.

To me, this type sounds more like `SetOnceAssertions`.

java/src/main/kotlin/io/spine/validation/java/setonce/SetOnceEnumField.kt
```diff
@@ -0,0 +1,99 @@
...
+        alterSetter()
+        alterEnumValueSetter()
+        alterBytesMerge(
+            currentValue = "${fieldName}_",
+            readerStartsWith = "${fieldName}_ = input.readEnum();"
+        )
+    }
+
+    /**
+     * Alters setter that accepts an enum constant.
```
armiol, line 65: It should be either
```
Alters a setter that accepts an enum constant.
```
or
```
Alters setters that accept an enum constant.
```

This comment applies to the similar cases as well.

yevhenii-nadtochii:
@alexander-yevsyukov @armiol PTAL

armiol (approved):
@yevhenii-nadtochii from my side, LGTM with a single comment.

PR #144, opened 2024-11-05
Migrate `(set_once)` tests to Kotlin

yevhenii-nadtochii:
This PR migrates recently added `(set_once)` tests to Kotlin.


yevhenii-nadtochii:

java-tests/validation-gen/src/test/kotlin/io/spine/test/tools/validate/setonce/SetOnceAssertions.kt
```diff
@@ -24,13 +24,27 @@
...
+import org.junit.jupiter.api.assertThrows
+
 /**
- * Contains tests for `(set_once)` field option.
+ * Assertions for `(set_once)` tests.
  */
-@CheckReturnValue
-@ParametersAreNonnullByDefault
-package io.spine.test.tools.validate.setonce;
+internal object SetOnceAssertions {
```
alexander-yevsyukov, line 36: Why do we need an object here? Why cannot we have just top level functions?

java-tests/validation-gen/src/test/kotlin/io/spine/test/tools/validate/setonce/SetOnceAssertions.kt
```diff
@@ -24,13 +24,27 @@
...
 
-import com.google.errorprone.annotations.CheckReturnValue;
+    /**
+     * Asserts that the given [runnable] throws [ValidationException].
+     */
+    fun assertValidationFails(runnable: () -> Unit) {
```
alexander-yevsyukov, line 41: This applies not only to `setonce`. So if we want these functions, they should probably go as test fixtures at the very start level of project dependencies.

java-tests/validation-gen/src/test/kotlin/io/spine/test/tools/validate/setonce/SetOnceAssertions.kt
```diff
@@ -24,13 +24,27 @@
...
+        assertThrows<ValidationException>(runnable)
+    }
 
-import javax.annotation.ParametersAreNonnullByDefault;
+    /**
+     * Asserts that the given [runnable] doesn't throw anything.
+     */
+    fun assertValidationPasses(runnable: () -> Unit) =
```
alexander-yevsyukov, line 48: Names of these functions are a bit long to be utilities. We also seem to have already things like `assertValid()` and `assertInvalid()`. Please have a look around.

java-tests/validation-gen/src/test/kotlin/io/spine/test/tools/validate/setonce/SetOnceAssertions.kt
```diff
@@ -24,13 +24,27 @@
...
+        assertDoesNotThrow(runnable)
```
alexander-yevsyukov, line 49: It is also useful to return a result of the executing the `runnable`. It helps when you need to verify the result of building a message. We also have tests that  run `.validate()` in combination with just building a message.

Again, please have a look around the code of the project.

alexander-yevsyukov: I mean the code like this:

```kotlin
val msg = assertDoesNotThrow {
    protoSet {
        element.add(toAny("42"))
        element.add(toAny(42))
    }
}
msg.validate().shouldBeEmpty()
```

alexander-yevsyukov: If we introduce something like `assertValid()` it may be implemented as the code above. But, again, we need to return the produced message so that the tests that use the utility can do something useful with it without the "gymnastics" with `var` declarations.

yevhenii-nadtochii: I've introduced them within an object as `(set_once)`-specific intentionally.

1. Assertions within `IsValid` are not applicable as they rely on validation being done in `builder()` method. The test subject here throws right from a builder setter.
2. Within `(set_once)` tests, we don't need the results, so I don't return them, allowing the test cases be one-lines.

I didn't look at them as general-purpose. I can move them back to `Spec*` files (two files need them), and make `private`. Or we can discuss vocally.

alexander-yevsyukov (approved):
LGTM, assuming agreements we've made in the video call.

PR #158, opened 2024-11-10
Fix package references in field type names

alexander-yevsyukov:
This PR fixes package references in codegen for `(set_once)` option. Previously, the package of the message type where a field is declared was used instead of the package of the message class of the field type.

Also the PR bumps codegen module dependencies to fix the scope of the call of a referenced field.

### DIFF
--- a/java/src/main/kotlin/io/spine/validation/java/setonce/SetOnceJavaConstraints.kt
+++ b/java/src/main/kotlin/io/spine/validation/java/setonce/SetOnceJavaConstraints.kt
@@ -32,16 +32,17 @@ import com.intellij.psi.PsiElementFactory
 import com.intellij.psi.PsiJavaFile
 import com.intellij.psi.PsiStatement
 import io.spine.protodata.ast.Field
+import io.spine.protodata.ast.toMessageType
 import io.spine.protodata.java.ClassName
 import io.spine.protodata.java.javaCase
 import io.spine.protodata.java.javaClassName
 import io.spine.protodata.java.render.findClass
 import io.spine.protodata.render.SourceFile
+import io.spine.protodata.type.TypeSystem
 import io.spine.string.camelCase
 import io.spine.tools.code.Java
 import io.spine.tools.psi.java.Environment.elementFactory
 import io.spine.tools.psi.java.execute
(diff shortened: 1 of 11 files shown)


java/src/main/kotlin/io/spine/validation/java/setonce/SetOnceJavaConstraints.kt
```diff
@@ -82,7 +83,8 @@ internal sealed class SetOnceJavaConstraints(
     protected val fieldGetterName = "get$fieldNameCamel"
     protected val fieldSetterName = "set$fieldNameCamel"
     protected val fieldGetter = "$fieldGetterName()"
-    protected val declaringMessage = declaredIn.message.javaClassName(declaredIn.fileHeader)
+    protected val declaringMessage =
+        field.declaringType.toMessageType(typeSystem).javaClassName(typeSystem)
```
yevhenii-nadtochii, line 87:
```suggestion
    protected val declaringMessage = field.declaringType
        .toMessageType(typeSystem)
        .javaClassName(typeSystem)
```

For me, it reads a bit easier when formatted this way.

java/src/main/kotlin/io/spine/validation/java/setonce/SetOnceMessageField.kt
```diff
@@ -50,9 +51,9 @@ internal class SetOnceMessageField(
         }
     }
 
-    private val fieldTypeClass = field.type.message
-        .javaClassName(declaredIn.fileHeader)
-        .canonical
+    private val fieldTypeClass =
+        field.type.message.toMessageType(typeSystem).javaClassName(typeSystem)
+            .canonical
```
yevhenii-nadtochii, line 56:
```suggestion
    private val fieldTypeClass = field.type.message
            .toMessageType(typeSystem)
            .javaClassName(typeSystem)
            .canonical
```

The same suggestion.

yevhenii-nadtochii (approved):
Good catch.

I will cover external message fields as I add negative tests cases.

alexander-yevsyukov:
It's the build of `core-java` who discovered the bug :)

PR #160, opened 2024-11-22
Implement tests for `(when)` option

yevhenii-nadtochii:
During the migration to typed expressions, we discovered an incompatible expression in `InTimeGenerator`. It was fixed in-place. This PR adds missing tests for `(when)` option and fixes another bug, which led to generation of validation code for Spine `Temporal` when the message declared `repeated google.protobuf.Timestamp`.

Now, there are two similar test suites to cover Protobuf `Timestamp` and Spine `Temporal` (particularly `io.spine.time.LocalDateTime`). Repeated fields have also been covered.


yevhenii-nadtochii:

java-tests/validating/src/test/kotlin/io/spine/test/options/when/ProtoTimestampWhenSpec.kt
```diff
@@ -0,0 +1,242 @@
...
+ * Protobuf [Duration] of fifty milliseconds.
+ *
+ * To shift the time into the past or future, we add or subtract a difference of this amount.
+ *
+ * There are two reasons for choosing fifty milliseconds:
+ *
+ * 1. The generated code uses `io.spine.base.Time.currentTime()` to get the current timestamp
+ * for comparison. In turn, this method relies on `io.spine.base.Time.SystemTimeProvider`
+ * by default, which has millisecond precision.
+ * 2. Adding too small amount of time to make the stamp denote "future" might be unreliable.
```
armiol, line 239: So why don't we choose two days instead?

java-tests/validating/src/test/kotlin/io/spine/test/options/when/SpineTemporalWhenSpec.kt
```diff
@@ -0,0 +1,243 @@
...
+ * Fifty milliseconds.
+ *
+ * To shift the time into the past or future, we add or subtract a difference of this amount.
+ *
+ * There are two reasons for choosing fifty milliseconds:
+ *
+ * 1. The generated code uses `io.spine.base.Time.currentTime()` to get the current timestamp
+ * for comparison. In turn, this method relies on `io.spine.base.Time.SystemTimeProvider`
+ * by default, which has millisecond precision.
+ * 2. Adding too small amount of time to make the stamp denote "future" might be unreliable.
```
armiol, line 240: Same question as the one above.

armiol (commented):
In general, LGTM. However, please see my question(s).

java-tests/validating/src/test/kotlin/io/spine/test/options/when/ProtoTimestampWhenSpec.kt
yevhenii-nadtochii: To be closer to the precision threshold. `two days` is already a seconds-based precision. Any value less than 1s would confirm millisecond precision during comparison.

alexander-yevsyukov (changes requested):
Please see my comments.

java-tests/validating/src/test/kotlin/io/spine/test/options/when/ProtoTimestampWhenSpec.kt
yevhenii-nadtochii: I've bumped to 500ms in case of slow VMs on CI.

armiol (approved):

alexander-yevsyukov (approved):
LGTM

Issue #67, opened 2022-10-30
Java runtime bundle contains unnecessary code


alexander-yevsyukov:
We, probably need to to ProGuard it. 11MB for such a thing is way too much.

armiol:
We have a complete control over what we put into this JAR. So instead of cutting it with ProGuard at a later stage, we can just not put redundant things into it in the first place. We already filter out some things like Gradle classes, at a time of assembly. 

Also, 11 MB seems not too much, if you ask me. That's the whole Spine framework (base, time, base-types, core, client, server) along with its dependencies.

In any case, current contents should be analyzed and tested in a non-Spine-library (!) project — otherwise the picture won't be clear. In this way, we'll be able to tell whether the types we carry on board are redundant or they are really needed.

alexander-yevsyukov:
Per #166, we don't need this artifact at all.

PR #166, opened 2024-11-29
Remove `java-runtime-bundle`, trim `java-bundle`

alexander-yevsyukov:
This PR removes the `java-runtime-bundle` module as it's not necessary, which eliminates the issue #67.

In relation to the dependency analysis performed for `java-runtime-bundle`, the size of `java-bundle` fat JAR was reduced by leaving only the code which is produced by the Validation library. The rest is available via ProtoData classpath.

## Other notable changes
 * Excluding transitive dependencies of the `java-bundle` module required introduction of remaining `local` dependency objects.
 * Properties of `Spine` and `ArtifactVersion` objects were deprecated in favour of new dependency objects.
 * Codegen dependencies were also bumped.


armiol (approved):
LGTM.

Issue #167, opened 2024-12-02
Extract common parts that generate code to create an instance of `ConstraintViolation`


yevhenii-nadtochii:
On the one hand we have `SetOnceConstraintViolation` and on the other `ErrorMessage.createViolation()` with `ErrorMessage.createCompositeViolation`. 

They are quite different, but should have the common base, which this issue suggests to extract. Another aspect of this issue it to create consistent instances of `ConstraintViolation` throughout the validation library.

The first implementation produces code that creates a `ConstraintViolation` for a specific option, for this reason it is relatively small. It uses our Expression API, and takes into account that the `ConstraintViolation` instance expects a `printf` string format along with a list of parameters for it.

The second implementation is coupled with `Rule`s-based validation codegen. It introduces own entities like: `ErrorMessages`, `ErrorMessage`, `Template`. And something similar to our Expression API from ProtoData: `Span`, `LiteralSpan`, `ExpressionSpan`. Though, it uses Java Poet to "partially" prepare the code. Another aspect is that it doesn't return a `printf` string, instead it interpolates parameters right into the format string.

So, in the scope of this issue, the following should be done:

1. Extract the common parts of `ContraintValidation` creation.
2. Re-use them in `SetOnceConstraintViolation` and `Rule`s codegen.
3. Migrate the remaining code around to PSI/Expression API.

yevhenii-nadtochii:
An instance of `ConstraintViolation` is created in a unified way now. Though, it doesn't apply to rules-based generation, as it is going to be removed.

PR #168, opened 2024-12-03
Implement codegen validation for `(goes)` option

yevhenii-nadtochii:
This PR implements codegen-based validation of `(goes)` option. 

In particular, it does the following:

1. Adds `GoesPolicy`, which emits the validation rule for the target field.
2. Adds positive tests for one-way and mutually dependent fields.
3. Removes `(goes)` from `Workaround.kt` file, and the runtime tests for this option.

The other aspects of `(goes)` feature are going to be addressed with separate PRs:

1. Error messages and tests for them – along with #167. There is some contradiction with how options treat `default_messages` and custom `error_msg`. It should be discussed and addressed.
2. Negative tests are to be implemented along with negative tests for `(set_once)`.
3. Add more details to the option docs.


yevhenii-nadtochii:
@armiol @alexander-yevsyukov please confirm specs for `(goes)` option. The present documentation doesn't say about the supported target (which declares the option) and companion (which is specified in `with`) field types.

<details>
<summary>The docs fragment from `options.proto`.</summary>

```
// Specifies that a message field can be present only if another field is present.
//
// Unlike the `required_field` that handles combination of required fields, this option is useful
// when it is needed to say that an optional field makes sense only when another optional field is
// present.
//
// Example: Requiring mutual presence of optional fields.
//
//    message ScheduledItem {
//        ...
//        spine.time.LocalDate date = 4;
//        spine.time.LocalTime time = 5 [(goes).with = "date"];
//    }
//
message GoesOption {

    // The default error message format string.
    //
    // The first parameter is the name of the field for which we specify the option.
    // The second parameter is the name of the field set in the "with" value.
    //
    option (default_message) = "The field `%s` can only be set when the field `%s` is defined.";

    // A name of the field required for presence of the field for which we set the option.
    string with = 1;

    // A user-defined validation error format message.
    string msg_format = 2 [deprecated = true];

    // A user-defined validation error format message.
    //
    // May include the token `{value}` for the actual value of the field. The token will be replaced
    // at runtime when the error is constructed.
    //
    string error_msg = 3;
}
```
</details>

In its docs, the option is compared to `required` and should likely mimic its behavior. So, I suggest the following list of supported target and companion fields:

1. Messages and enums. Default instances are considered to be non-set.
2. Repeated fields and maps. Empty collections are non-set fields.
3. `string` and `bytes`. An empty string and empty byte array are non-set.

The same field types are supported by `(required)`. 

Numeric and boolean fields are not intended to be either target fields or companions, as it is impossible to determine whether such a field was set or not. The `(required)` option also treats these fields as always set because their default values are quite "normal". The remaining question is whether we should silently ignore these fields when used with the option (as `(required)` does) or throw an error instead.

armiol:
@yevhenii-nadtochii Yes, I would consider it to be like a conditional `required`. With the applications exactly like `required` has.

java-tests/validating/src/test/kotlin/io/spine/test/options/goes/TestDataMutual.kt
```diff
@@ -0,0 +1,117 @@
...
+import io.spine.test.tools.validate.MutualStringCompanion
+import kotlin.reflect.KClass
+import org.junit.jupiter.api.Named.named
+import org.junit.jupiter.params.provider.Arguments.arguments
+
+/**
+ * Provides data for parameterized [GoesMutualITest].
+ */
+@Suppress("unused") // Data provider for parameterized test.
+internal object TestDataMutual {
```
armiol, line 47: I would move it to `.given` package and rename to `...Env`.

java-tests/validating/src/test/kotlin/io/spine/test/options/goes/TestDataOneWay.kt
```diff
@@ -0,0 +1,118 @@
...
+import io.spine.test.tools.validate.StringCompanion
+import kotlin.reflect.KClass
+import org.junit.jupiter.api.Named.named
+import org.junit.jupiter.params.provider.Arguments.arguments
+
+/**
+ * Provides data for parameterized [GoesOneWayITest].
+ */
+@Suppress("unused") // Data provider for parameterized test.
+internal object TestDataOneWay {
```
armiol, line 47: Same as above.

armiol (changes requested):
@yevhenii-nadtochii mostly LGTM. See my comments, though.

yevhenii-nadtochii:
@armiol PTAL

armiol (approved):

PR #171, opened 2024-12-16
Adapt the library to `RangeOption`

yevhenii-nadtochii:
This PR adapts the library to the latest changes in `options.proto`. 

In particular, it does the following:

1. Takes the latest versions on `base`, `base-types` and ProtoData.
2. Addresses migration of `range` from `string` to `RangeOption` type.
3. Addresses rephrased error messages.
4. Makes `set_once` handle the updated placeholder names.
5. Removes some unused Proto imports and fixes the build file deprecations.

Please note, this PR makes `validation` buildable again. Full support of all placeholders for all options will be added in the upcoming PR (most likely here: #169). The same applies to `SetOnceViolationConstraint`, code of which should become shared with `Rule`-based validation.


yevhenii-nadtochii:

armiol (approved):

PR #179, opened 2025-01-04
Fix publishing

alexander-yevsyukov:
This PR bumps the version number to pass the publishing. The version was increased by mistake for the [previous PR](#178), which [failed the publishing](https://github.com/SpineEventEngine/validation/actions/runs/12612662406/job/35149645272). 

The `checkVersionIncrement` tasks did not guard against the non-increasing the version because it did not take into account custom artifact prefix (`spine-validation-`) which is used for the modules of this project. This PR improves the `CheckVersionIncrement` task in this regard, and the change would be propagated to `config` shortly.


alexander-yevsyukov:
@armiol, PTAL.

armiol (approved):

PR #180, opened 2025-01-07
Remove nested violations

yevhenii-nadtochii:
This PR does the following:

1. Removes usages of the deprecated `(if_invalid)` option. There's no error message for `(validate)` anymore.
2. Deprecates nested violations and removes its usages. We no longer create an umbrella violation for `(validate)`.

I haven't introduced a nullable error message in `BoolFieldOptionView` for two reasons:

1. An empty error message can not be misused. `(valiate)` has its own `CodeGenerator`.
2. We would have to tackle with nullability down the `Rule`'s framework, which is quite useless because `Rule`s are Protobuf messages, where `no string` remains an `empty string`.


yevhenii-nadtochii:

model/src/main/kotlin/io/spine/validation/ValidatedFieldView.kt
```diff
@@ -29,38 +29,35 @@ package io.spine.validation
...
     override fun enableValidation() {
         builder()!!.setValidate(true)
     }
 
-    @Subscribe
-    override fun onErrorMessage(
-        @External @Where(field = OPTION_NAME, equals = IF_INVALID) e: FieldOptionDiscovered
-    ) = super.onErrorMessage(e)
+    override fun extractErrorMessage(option: Option): String = error(
+        "Can not extract custom error message for `($VALIDATE)` option using `$option`. " +
```
armiol, line 54: "Cannot".

proto/context/src/main/proto/spine/validation/field_views.proto
```diff
@@ -30,11 +30,8 @@ message ValidatedField {
 
     FieldId id = 1;
 
-    // Message appearing when this field is not valid.
-    string error_message = 2;
```
armiol, line 34: Given this is a part of the public API, I would deprecate the field first.

java-tests/validating/src/test/kotlin/io/spine/test/options/ValidateConstraintTest.kt
```diff
@@ -61,26 +61,20 @@ internal class ValidateConstraintTest {
 
         error.shouldBePresent()
 
-        val violations = error.get().constraintViolationList
-
-        // We should have 2 violations instead of 1.
+        // We should have 3 violations instead of 2.
```
armiol, line 64: But below the code says `violations.size shouldBe 2`. Who's right?

model/src/main/kotlin/io/spine/validation/ValidatedFieldView.kt
```diff
@@ -29,38 +29,35 @@ package io.spine.validation
...
-import io.spine.protobuf.unpack
 import io.spine.protodata.ast.Option
 import io.spine.protodata.ast.event.FieldOptionDiscovered
 
 /**
- * A view of a field that is marked with `validate`.
+ * A view of a field that is marked with `(validate)`.
+ *
+ * Note: this option [does not have][NO_ERROR_MESSAGE] an error message. It triggers
+ * in-depth validation and, in case of errors, propagates only the initial cause.
```
armiol, line 39: "...propagates only the first one occurred."

model/src/main/kotlin/io/spine/validation/ValidatedFieldView.kt
```diff
@@ -29,38 +29,35 @@ package io.spine.validation
...
         builder()!!.setValidate(true)
     }
 
-    @Subscribe
-    override fun onErrorMessage(
-        @External @Where(field = OPTION_NAME, equals = IF_INVALID) e: FieldOptionDiscovered
-    ) = super.onErrorMessage(e)
+    override fun extractErrorMessage(option: Option): String = error(
+        "Can not extract custom error message for `($VALIDATE)` option using `$option`. " +
+                "`($VALIDATE)` does not support custom error message."
```
armiol, line 55: "... does not support custom error messages". Alternatively, if left singular, you will need an article.

armiol (changes requested):

alexander-yevsyukov (changes requested):
Please see my comments.

alexander-yevsyukov (approved):
LGTM

PR #183, opened 2025-01-10
Fill in placeholder values for `TemplateString`

yevhenii-nadtochii:


yevhenii-nadtochii:
Closing this PR as no longer needed.

It is very hard to add placeholders support with `Rule`-based constraints rendering. We decided to have dedicated `CodeGenerator` for each option explicitly. They will re-use some pieces of code, but each generator will know which option it generates, making it easier to bring option-specific functionality.

PR #185, opened 2025-01-16
Refactor `JavaValidationRenderer`

yevhenii-nadtochii:
This PR refactors `JavaValidationRenderer` and suggests renaming of `MessageValidation` → `CompilationMessage`. I need this change to proceed with the migration to PSI and simultaneous support of rules and standalone generators (as we complete the migration to the latter).

Changes to `JavaValidationRenderer`:

1. Revises the class top-level documentation. It was outdated.
2. Decreases the number of intermediate and file-level methods in the file.
3. The top-level documentation is directly mapped to the body of `render()` method now. A significant part of this method is local comments, describing why all modifications of a message file cannot be done under a single `forEach`. I describe this inconsistency, but leave it "as is" because I'm going to migrate to PSI in the upcoming PRs.
4. I removed the `protected` modifier from `render()` because it is unnecessary with `override`, and makes the entry point a bit harder to locate.

Other changes:

1. I have removed `Validations` collection, we can easily go without this class. We already have many classes with "Validation" in their names. Takes time to differentiate their purposes.
2. Moved declaration of `FieldId` up in `field_views.proto` file.
3. Renamed `MessageValidation` to `CompilationMessage`. 
  3.1. This projection stores views for each message being compiled.
  3.2. Having one or more constraints is optional. We always add `validate()` to all messages (and related stuff), which go through the library no matter if they have constraints or not.
  3.3. When we complete migration from `Rule`, this class will disappear completely. Each generator will query only a view of interest, like `RequiredFieldView` or `DistinctFieldView`.
  3.4. Its usages talked more about a `compilationMessage`, and less as of a `set of constraints`.
  3.5. This eliminates another class with "Validation" in its name.

Unrelated changes:

1. Renamed a local variable in `SetOnceValidationRenderer`. Both options are good enough, but the new one looks a bit better to me. And `JavaValidationRenderer` also uses `allCompilationMessages`.
2. Took the latest ProtoData.


yevhenii-nadtochii:

proto/context/src/main/proto/spine/validation/compiled_message.proto
```diff
@@ -32,17 +32,17 @@ import "spine/options.proto";
...
+option java_outer_classname = "CompiledMessageProto";
 option java_multiple_files = true;
 
 import "spine/protodata/ast.proto";
 import "spine/validation/rule.proto";
 
-// All the validation rules of a single message type.
-message MessageValidation {
+// A message compiled by the validation library with its constraints, if any.
+message CompiledMessage {
```
armiol, line 42: As long as in Spine there is a "Model Compiler" (which in turn heavily uses the Validation codegen capabilities), we cannot use `CompiledMessage`, because it clashes with that domain. Until I saw the before-after in this changeset, I had no idea what a `CompiledMessage` was. My only guess was that it was something describing the work of the Model Compiler routines completed for some particular `Message` instance, more of an item in some audit log. 

I agree that `MessageValidation` isn't the best name. But still, what this declaration represents is by no means a result of a compilation process in that sense which we assume in Model Compiler. That's why it cannot be "compiled" (a participle, meaning "it's done"), It's rather an intermediate view of a "to-do" list for the next stages.

java/src/main/kotlin/io/spine/validation/java/JavaValidationRenderer.kt
```diff
@@ -52,86 +51,75 @@ import io.spine.validation.java.point.ValidateBeforeReturn
...
- * performs the given action for each file.
- */
-private fun SourceFileSet.forEachSourceFile(
-    messageTypes: Set<MessageWithFile>,
-    action: SourceFile<Java>.() -> Unit
-) {
-    messageTypes
-        .map { m -> m.message.javaFile(m.fileHeader) }
-        .mapNotNull { path -> find(path) }
-        .distinct()
```
armiol, line 119: Are you sure that the newly introduced `distinct()` (line 98) is semantically equal to this whole block? I suspect there might be a subtle difference.

java/src/main/kotlin/io/spine/validation/java/ValidationConstraintsCode.kt
```diff
@@ -126,7 +126,7 @@ internal class ValidationConstraintsCode private constructor(
         /**
          * Creates a new instance with the generated validation constraints code.
          */
-        fun generate(r: JavaValidationRenderer, v: MessageValidation): ValidationConstraintsCode {
+        fun generate(r: JavaValidationRenderer, v: CompiledMessage): ValidationConstraintsCode {
```
armiol, line 129: As a general comment: when renaming something, please check the corresponding variable names, too.

proto/context/src/main/proto/spine/validation/field_views.proto
```diff
@@ -11,17 +11,30 @@ option java_multiple_files = true;
...
 
     FieldId id = 1;
 
-    // Message appearing when this required field is not set.
+    // The error message template.
     string error_message = 2;
 
     // `true` if the field is marked with `(required) = true`, `false` otherwise.
-    bool required = 3;
+    bool enabled = 3;
```
armiol, line 34: The field was called `required` because it's written like that in the consumer code.

"Enabled" is about something else. "Whether the validation is enabled or not?"

proto/context/src/main/proto/spine/validation/field_views.proto
```diff
@@ -11,17 +11,30 @@ option java_multiple_files = true;
 
 import "spine/protodata/ast.proto";
 
+// Uniquely identifies a message field.
```
armiol, line 14: I would say it uniquely identifies a declaration of a field in some `Message`. 

It doesn't identify the field instance itself, as one might think reading this.

proto/context/src/main/proto/spine/validation/field_views.proto
```diff
@@ -11,17 +11,30 @@ option java_multiple_files = true;
...
+message FieldId {
+
+    // Name of the containing type.
+    protodata.TypeName type = 1;
+
+    // Short name of the field.
```
armiol, line 20: I think it's enough that we call the field name "short" in the declaration of the `FieldName` itself.

proto/context/src/main/proto/spine/validation/field_views.proto
```diff
@@ -37,26 +50,16 @@ message ValidatedField {
...
 
     FieldId id = 1;
 
-    // Message appearing when this required field is not set.
+    // The error message template.
     string error_message = 2;
 
-    // Value of `set_once` boolean option.
+    // `true` if the field is marked with `(set_once) = true`, `false` otherwise.
     bool enabled = 3;
```
armiol, line 63: Basically, I would have this named `set_once` for that same reason as above.

proto/context/src/main/proto/spine/validation/compiled_message.proto
```diff
@@ -55,7 +55,7 @@ message MessageValidation {
 
     // Validation rules.
     //
-    // All these rules must be met in order for a message to be regarded as valid.
```
armiol, line 58: Why remove? The meaning is somewhat changed, don't you think?

proto/context/src/main/proto/spine/validation/compiled_message.proto
armiol: Let's think of a better name. I will, too.

armiol (changes requested):
@yevhenii-nadtochii very dense PR, good job! However, in some places I think you went too far. Please see my comments.

proto/context/src/main/proto/spine/validation/compiled_message.proto
yevhenii-nadtochii: I thought about `CompiledMessage` as a message that is being compiled **right now**. It can also be `CompilingMessage`, `CompilationMessage`.

java/src/main/kotlin/io/spine/validation/java/JavaValidationRenderer.kt
yevhenii-nadtochii: Now, ProtoData combines calls of `javaFile(header)`, `find(path)` and unchecked cast to `SourceFile<Java>` in a single `SourceFileSet.javaFileOf()` method.

This new method throws when `find(path)` returns `null` (meaning no Java file). I have added ` if (!sources.hasJavaRoot) { return }` in the beginning of the renderer to avoid this.

proto/context/src/main/proto/spine/validation/field_views.proto
yevhenii-nadtochii: It is a bit weird both ways. 

I suggest not create `RequiredField` view for non-required fields at all. The same with the rest of such views. We do nothing when such a flag is `false` because `RequiredField` with `required = false` is `NonRequiredField`.

Having this `bool something = false` flag in `SomethingField` means this `something field` is not `something`.

proto/context/src/main/proto/spine/validation/field_views.proto
yevhenii-nadtochii: >> It doesn't identify the field instance itself, as one might think reading this.

It does because it has both the parent type name and the field name.

yevhenii-nadtochii:
@armiol PTAL

proto/context/src/main/proto/spine/validation/field_views.proto
```diff
@@ -11,17 +11,30 @@ option java_multiple_files = true;
 
 import "spine/protodata/ast.proto";
 
+// Uniquely identifies a field declaration within the compilation messages.
```
armiol, line 14: I think this type addresses not only "compilation" messages. It's a part of AST, so I say we shouldn't be specific about "compilation" messages.

armiol (approved):
@yevhenii-nadtochii LGTM with the respect of the comment.

PR #186, opened 2025-01-22
Migrate `ValidationCode` to PSI

yevhenii-nadtochii:
This change set migrates the main Java validation renderer to PSI.

All Java class modifications are now performed within a single method in `MessageValidationCode.render(psiFile)`, allowing for quick grasp of what is actually mutated and how.

This class deals with JavaPoet only in a degree to remain compatible with rules-based `CodeGenerator` output. These generators produce `CodeBlock`, `MethodSpec` and `FieldSpec`, which are easily converted to `PsiStatements` (our custom wrapper for `PsiCodeBlock`, see below), `PsiMethod` and `PsiField`.

**Note:** this PR doesn't bring new features, only migration to PSI and refactoring. Another preparational step to introduce option-specific generators.

## Particular changes

1. Explicitly mentioned in docs to `@Validated` and `@NonValidated` that these annotations are not meant for runtime.
2. Added tests ensuring the generated messages implement `ValidatableMessage` and `ValidatingBuilder` interfaces.
3. `CodeGenerator.supportingMemebers()` has been split into `supportingFields()` and `supportingMethods()`. We can't create general-purpose elements with PSI, only specific ones. Usage of `Queriyng` were replaced with `TypeSystem`.
4. I've made `ValidationPlugin` abstract, with `renderers` being expected on the constructor. This makes the declaration of `JavaValidationPlugin` more straightforward.
5. `MessageValidationCode` now contains both modifications of the message class and its builder. I prefer having them both in the same method because they are intertwined. Modifying of `Builder.build()` doesn't make sense without presence of `Message.validate()`.
6. Now, we generate only `return java.util.Optional.empty();` in `validate()` when the message doesn't declare constraints. No need to put there an `if` block, which always evaluates to `Optional.empty()`.
7. Renamed `SetOnceValidationRenderer` to `SetOnceRenderer`. Now, it looks like this `Validation` word is redundant.

## Removed classes

These classes were merged into a single `MessageValidationCode`:

1. `ImplementValidatingBuilder`.
2. `ValidateMethodCode`.
4. `ValidationCode`.
5. `ValidationConstraintsCode`.

These classes are no longer needed (all from `java.point` package):

1. `BuilderInsertionPoint`
2. `BuilderMethodReturnTypeAnnotation`
4. `BuildPartialReturnTypeAnnotation`.
5. `CachingJavaClassSource`.
6. `ExtraValidation`.
7. `ParsedSources`.
8. `ValidateBeforeReturn`. 
9. `BuilderMethodsSpec` was testing the insertion points, not the generated code.
10. `MockPrinter` was a fixture for `BuilderMethodsSpec`.

## API suggestions for PSI

1. `PsiClass.nested(simpleName: String): PsiClass` – allows easily get a nested `Message.Builder` when you already have a `PsiClass` of the `Message` at your disposal.
2. `PsiElement.findFirstByText(...)`, which accepts `String`. I have taken it from `(set_once)` implementation because this is the most reliable way to find a place to inject the code somewhere in the method body. Even if we know, that we want to inject before the last statement – it doesn't help. The last `PsiElement` in `PsiMethod.body` is always `right brace`. Ok, let's try its previous sibling (should it be `return`?) – no, it can be `return ...;`, `whitespace` or even `line break`. Depends on the formatting. PSI stores formatting elements along with other elements.
3. `PsiStatements` wrapper to avoid dealing with `PsiCodeBlock` when we need to represent a list of statements. In `validation` codebase, we treat code blocks as `List<Statement>`. But within PSI, a code block is more about Java grammar, where it is a list of statements **and the curly braces** `{ <statements> }`. This sophisticates 1) creating a set of statements from text and 2) inserting them to other code blocks.
4. `PsiClass.findMethodBySignature()` and `PsiClass.methodWithSignature()`, which were introduced for `(set_once)`.


yevhenii-nadtochii:

java/src/main/kotlin/io/spine/validation/java/JavaValidationPlugin.kt
```diff
@@ -26,48 +26,21 @@
...
- * A plugin that sets up everything needed to generate Java validation code.
+ * An implementation of [ValidationPlugin] for Java language.
  *
- * This plugin uses a delegate plugin to set up some of the components needed for
- * code generation. By default, a [ValidationPlugin] is used. However, API users may
- * extend this plugin's behavior and supply a more rich base plugin.
- *
- * @param base The base plugin to extend.
- * @constructor Creates an instance that contains all the components from the given [base] plugin.
+ * Note: [`(set_once)`][SetOnceRenderer] option is rendered with its own
```
alexander-yevsyukov, line 35: It's not the option which is rendered. We render validation code serving the `(set_once)` constraint.

I would start the explanation of the matter with the enumeration of the renderers included into the `ValidationPlugin`, and then explain them both briefly.

java/src/main/kotlin/io/spine/validation/java/MessageValidationCode.kt
```diff
@@ -0,0 +1,226 @@
...
+    private fun MessagePsiClass.declareValidateMethod() {
+        // `CodeBlock` is printed with a new line in the end, so no need to call `joinByLine()`.
+        // And we have to trim because we don't need a trailing empty line.
+        val formattedConstraints = constraints.joinToString(separator = "").trim()
+        val psiMethod = elementFactory.createMethodFromText(
+            """
+            |public java.util.Optional<io.spine.validate.ValidationError> validate() {
+            |    ${validateMethodBody(formattedConstraints)}
+            |}
+            """.trimMargin(), this
```
alexander-yevsyukov, line 155: Why do we use the style which requires `trimMargin()` here, and the `trimIndent()` style below?

java/src/main/kotlin/io/spine/validation/java/JavaValidationPlugin.kt
```diff
@@ -26,48 +26,21 @@
...
- * This plugin uses a delegate plugin to set up some of the components needed for
- * code generation. By default, a [ValidationPlugin] is used. However, API users may
- * extend this plugin's behavior and supply a more rich base plugin.
- *
- * @param base The base plugin to extend.
- * @constructor Creates an instance that contains all the components from the given [base] plugin.
+ * Note: [`(set_once)`][SetOnceRenderer] validation code is rendered with its own
+ * renderer because it significantly differs from the rest of constraints. This option
+ * modifies the message builder behavior. The rest of the constraints are more about
+ * message state assertions.
```
armiol, line 38: Did you mean "builder state assertions"?

java/src/main/kotlin/io/spine/validation/java/JavaValidationRenderer.kt
```diff
@@ -26,152 +26,35 @@
...
- * Also, it puts the following annotations:
- *
- * 1. [Validated] for [Message.Builder.build] method.
- * 2. [NonValidated] for [Message.Builder.buildPartial] method.
- *
- * Note: there is also [ImplementValidatingBuilder] renderer that makes [Message.Builder]
- * implement [io.spine.validate.ValidatingBuilder] interface. Logically, it is a part
- * of this renderer, but as for now, it is a standalone renderer before this class
- * migrates to PSI.
+ * This rendered is applied to every [Message] being compiled, even if the message
```
armiol, line 39: "This rendered ..." has some grammar issues.

java/src/main/kotlin/io/spine/validation/java/MessageValidationCode.kt
```diff
@@ -0,0 +1,237 @@
+/*
+ * Copyright 2024, TeamDev. All rights reserved.
```
armiol, line 2: It's 2025.

java/src/main/kotlin/io/spine/validation/java/MessageValidationCode.kt
```diff
@@ -0,0 +1,237 @@
...
+
+/**
+ * Generates validation code for the given [CompilationMessage].
+ *
+ * This class modifies both the generated Java class of the message and its builder.
+ *
+ * All validation logic is injected into the [ValidatableMessage.validate] method
+ * within the message class itself. This allows a message instance to be validated even
+ * after it has been built. Note that this method does not throw exceptions directly;
+ * instead, it returns any detected violations, if any.
```
armiol, line 70: "...it returns the detected violations, if any."

java/src/main/kotlin/io/spine/validation/java/MessageValidationCode.kt
```diff
@@ -0,0 +1,237 @@
...
+ *
+ * The message builder is modified to invoke the [ValidatableMessage.validate] just
+ * before returning from its [build][com.google.protobuf.Message.Builder.build] method.
```
armiol, line 73: "before returning the result from ..."

java/src/main/kotlin/io/spine/validation/java/MessageValidationCode.kt
```diff
@@ -0,0 +1,237 @@
...
+                implementValidatingBuilder()
+                injectValidationIntoBuildMethod()
+                annotateBuildReturnType()
+                annotateBuildPartialReturnType()
+            }
+        }
+    }
+
+    /**
+     * Generates a Java [constraint][constraints], [supportingFields] and [supportingMethods]
```
armiol, line 115: "a" cannot be used with plurals. Please review.

java/src/main/kotlin/io/spine/validation/java/MessageValidationCode.kt
```diff
@@ -0,0 +1,237 @@
...
+
+    /**
+     * Declares `validate()` method in this [MessagePsiClass].
+     *
+     * This method does the following to address the formatting issues:
+     *
+     * 1. [constraints] are joined to [String] using [joinToString] instead of `joinByLine()`
+     * because the contained code blocks already have new lines when converted to string.
+     * 2. For the same reason, we trim the result because a trailing empty line is not needed.
+     * 3. When creating PSI method from the text, we have to use [trimMargin] and '|' symbols
```
armiol, line 154: Please review this p.3 using any grammar checker. I personally use Grazie (as a Chrome extension).

I suspect some prepositions are missing.

java/src/main/kotlin/io/spine/validation/java/psi/PsiClassExts.kt
```diff
@@ -24,14 +24,27 @@
...
 
 import com.intellij.psi.PsiClass
 import com.intellij.psi.PsiMethod
 import io.spine.tools.psi.java.Environment.elementFactory
 
 /**
- * Searches for a method in this [PsiClass] matching the given signature
+ * Returns a nested class declared in this [PsiClass].
+ *
+ * @param simpleName The class simple name.
```
armiol, line 36: "The simple name of the class."

model/src/main/kotlin/io/spine/validation/ValidationPlugin.kt
```diff
@@ -29,16 +29,22 @@ package io.spine.validation
...
 import io.spine.validation.required.RequiredIdOptionPolicy
 import io.spine.validation.required.RequiredIdPatternPolicy
 import io.spine.validation.required.RequiredPolicy
 
 /**
- * A ProtoData plugin which attaches validation-related policies and views.
+ * The basic implementation of ProtoData validation plugin, which builds
+ * language-agnostic representation of the declared constraints.
+ *
+ * The concrete implementations should provide [renderers], which will
```
armiol, line 42: I would remove "will".

armiol (changes requested):
@yevhenii-nadtochii LGTM in general. Please see my comments.

yevhenii-nadtochii:

armiol (approved):

alexander-yevsyukov (approved):
LGTM

PR #188, opened 2025-01-30
Generate code for `(required)` with a standalone generator

yevhenii-nadtochii:
This PR starts migration from `Rule`-based generators to option-specific ones.

The ground interface for such generators is `OptionGenerator`. The first implementation is `RequiredOption`, which generates code for fields marked with `(required)`. Now, `RequiredPolicy` does the job related to checks about whether the option is applied correctly. Its `RequiredView` is just a plain view.

Temporarily, `MessageValidationCode` handles both option-specific generators and rule-based ones. Further, we would like to remain only the former.

Other changes are the following:

1. Added `validate(FieldPath)` overloading, allowing option-specific generators set `ConstraintViolation.fieldPath` with the correct value when an option is invoked within `(validate)` constraint. `validate()` just invokes the private overloading with the default instance of path. All invocations within `(validate)` should bypass `parent+this_field` value.
2. Extracted creation of `ConstraintViolation` and `TemplateString` expressions to methods as they should be the same for all options.
3. Picked up changes from the latest `ProtoData`, `tool-base` and `base`.


yevhenii-nadtochii:
@armiol @alexander-yevsyukov Please take a look at this draft.

It does not build because it requires [changes](https://github.com/SpineEventEngine/ProtoData/pull/318) to ProtoData. Have a look, let's discuss what should go to ProtoData, and then I will make this PR build.

armiol:
@yevhenii-nadtochii is it still a draft?

java/src/main/kotlin/io/spine/validation/java/RequiredOptionGenerator.kt
```diff
@@ -0,0 +1,150 @@
...
+            .getter<Any>()
+        val message = view.errorMessage
+        return CodeBlock(
+            """
+            if ($getter.equals(${defaultValue(field)})) {
+                var fieldPath = ${fieldPath(field.name.value, parent)};
+                var violation = ${violation(field, ReadVar("fieldPath"), message)};
+                $violations.add(violation);
+            }
+            """.trimIndent()
```
alexander-yevsyukov, line 99: Why do we have `trimIndent()` here?

java/src/main/kotlin/io/spine/validation/java/RequiredOptionGenerator.kt
```diff
@@ -0,0 +1,150 @@
...
+    }
+
+    /**
+     * Determines the value for each of the supported `(required)` placeholders.
+     */
+    private fun supportedPlaceholders(
+        field: Field,
+        path: Expression<FieldPath>
+    ): Map<ErrorPlaceholder, Expression<String>> {
+        val pathAsString = ClassName("io.spine.base", "FieldPaths")
```
alexander-yevsyukov, line 142: Why don't we use the actual class reference instead of strings here?

java/src/main/kotlin/io/spine/validation/java/protodata/CodeBlock.kt
```diff
@@ -0,0 +1,57 @@
...
+ * ```
+ * CodeBlock(
+ *     """
+ *     if (fieldValue.equals("hello")) {
+ *         var fieldPath = getFieldPath();
+ *         var violation = createViolation(fieldPath);
+ *         violations.add(violation);
+ *     }
+ *     return violations;
+ *     """.trimIndent()
```
alexander-yevsyukov, line 47: Again, why not `trimMargin()`?

alexander-yevsyukov (commented):

java/src/main/kotlin/io/spine/validation/java/protodata/CodeBlock.kt
armiol, line 47: I would build any trimming into the functionality of the class so that end-users won't have to do that every time.

java/src/main/kotlin/io/spine/validation/java/protodata/CodeBlock.kt
```diff
@@ -0,0 +1,57 @@
...
+ */
+
+package io.spine.validation.java.protodata
+
+import io.spine.protodata.java.AnElement
+import io.spine.protodata.java.Statement
+import io.spine.string.joinByLines
+
+/**
+ * Represents one or more Java statements.
```
armiol, line 34: How do we build a block of some code with the comments? E.g.
```
val field = ...
// Here is some explanation.
field.doSomething();
```

java/src/main/kotlin/io/spine/validation/java/protodata/MemberDeclaration.kt
```diff
@@ -24,12 +24,13 @@
...
-package io.spine.validation.required
+package io.spine.validation.java.protodata
 
-import io.spine.validation.BoolFieldOptionRepo
-import io.spine.validation.RequiredField
+import io.spine.protodata.java.AnElement
 
 /**
- * A repository for the [RequiredFieldView]s.
+ * Represents class level member declaration.
```
armiol, line 32: "Represents a class-level declaration of a member."?

proto/context/src/main/proto/spine/validation/field_views.proto
```diff
@@ -30,11 +30,8 @@ message RequiredField {
     // The error message template.
     string error_message = 2;
 
-    // `true` if the field is marked with `(required) = true`, `false` otherwise.
-    bool required = 3;
```
armiol, line 34: If you delete this field, let's modify the documentation of its parent message to reflect the difference in the behavior. So from now on, the instances of `RequiredField` are only created for the declarations marked with `(required) = true`.

java-tests/validating/src/test/kotlin/io/spine/test/options/required/RequiredRepeatedEnumITest.kt
```diff
@@ -41,7 +41,7 @@ internal class RequiredRepeatedEnumITest {
     @Test
     fun `require at least one item`() {
         val instance = Collections.newBuilder()
-        assertViolation(instance, field, "must not be empty")
+        assertViolation(instance, field, "must have a value")
```
armiol, line 44: This is weird. Collection has a value. It's just empty.

java/src/main/kotlin/io/spine/validation/java/MessageValidationCode.kt
```diff
@@ -72,11 +74,18 @@ private typealias BuilderPsiClass = PsiClass
  * The message builder is modified to invoke the [ValidatableMessage.validate] just before
  * returning the result from its [build][com.google.protobuf.Message.Builder.build] method.
  * If one or more violations are detected, the builder will throw an exception.
+ *
+ * Notice on multiline string literals in this class: when creating a PSI method from text,
+ * we prefer using [trimMargin] with '|' symbols (the default separator) to explicitly
+ * specify the minimal common indent. For some reason, [trimIndent] cannot always correctly
+ * detect the expected common indent. It may distort the resulting formatting, often leading
```
armiol, line 81: Why is this a problem giving we reformat the code afterwards?

java/src/main/kotlin/io/spine/validation/java/MessageValidationCode.kt
```diff
@@ -117,11 +127,15 @@ internal class MessageValidationCode(
      * A single rule always generates a single Java constraint represented by a [CodeBlock].
      * The number of supported members is not restricted. A single rule may generate zero,
      * one, or more supporting members.
+     *
+     * Note: [CodeGenerator.code] returns JavaPoet's code block, which we convert
+     * to [CodeBlock] from ProtoData Expression API, so that they could be added
```
armiol, line 132: Who are these "they"?

java/src/main/kotlin/io/spine/validation/java/MessageValidationCode.kt
```diff
@@ -117,11 +127,15 @@ internal class MessageValidationCode(
...
+     * to a single collection with the ones produced by [generators].
```
armiol, line 133: "....with those produced..."

java/src/main/kotlin/io/spine/validation/java/MessageValidationCode.kt
```diff
@@ -145,46 +159,93 @@ internal class MessageValidationCode(
...
     }
 
-    private fun validateMethodBody(formattedConstraints: String): String =
-        if (formattedConstraints.isEmpty()) {
-            "return java.util.Optional.empty();"
-        } else {
+    /**
+     * Declares `validate(FieldPath)` method in this [MessagePsiClass].
+     *
+     * All actual validations are rendered in this `private` overloading. It accepts
```
armiol, line 183: I don't understand what this and the next paragraph mean. Something is missing. And "validations" is not a thing.

java-tests/validating/src/test/kotlin/io/spine/test/options/ValidateConstraintTest.kt
```diff
@@ -67,8 +68,8 @@ internal class ValidateConstraintTest {
         val violations = error.get().constraintViolationList
         violations.size shouldBe 2
 
-        violations[0].fieldPath.getFieldName(0) shouldBe "first_line"
-        violations[1].fieldPath.getFieldName(0) shouldBe "town"
+        val failedFields = violations.map { it.fieldPath.getFieldName(0) }
+        failedFields shouldContainExactlyInAnyOrder listOf("first_line", "town")
```
yevhenii-nadtochii, line 72: This is because `OptionGenerator`-provided `if`s are rendered AFTER the `Rule`-provided ones in `validate(...)`.
Thus, affecting the ordering of violations. 

The emerged question is the following: do we provide ordering expectations regarding the produced violations? So, if fields with ordinal numbers `#4, #8, #9` fail, we get violations for `#4, #8, #9` in the same order.

For example, `Rule` framework renders constraints according to the order of declaration in the original .proto file. So, if `myField` is declared first and has three validation options, the first three `if` blocks in `validate(...)` will be asserting exactly the value of `myField()` getter. As for now, it is not true for `OptionGenerator`.

yevhenii-nadtochii:

java-tests/validating/src/test/kotlin/io/spine/test/options/ValidateConstraintTest.kt
armiol: @yevhenii-nadtochii we have briefly discussed the matter locally, and we think the ordering of violations does not matter. At least, as of now. We may want to tweak it in the future, but for now it does not matter.

java/src/main/kotlin/io/spine/validation/java/ConstraintViolations.kt
```diff
@@ -0,0 +1,95 @@
...
+ * @param errorMessage The error message template string.
+ * @param declaringType The message type being validated.
+ * @param fieldPath The path to the field containing an invalid value.
+ * @param fieldValue The field value, if any.
+ */
+public fun constraintViolation(
+    errorMessage: Expression<TemplateString>,
+    declaringType: TypeName,
+    fieldPath: Expression<FieldPath>,
+    fieldValue: Expression<*>? = null
```
armiol, line 55: Please have some details on when this parameter may be `null`. It's somewhat strange to see `null`s in the values of fields in a Proto message.

model/src/main/kotlin/io/spine/validation/required/RequiredPolicy.kt
```diff
@@ -26,83 +26,103 @@
...
  *
- * If any of these conditions are not met, nothing happens.
+ * 1. The field type is supported by the option.
+ * 2. The option value is `true`.
+ *
+ * If (1) is violated, the policy reports a compilation error.
+ *
+ * Violation of (2) means that the `(required)` option is applied correctly,
+ * but disabled. In this case, the policy emits [NoReaction] because we actually
+ * have a non-required field, marked with `(required)`.
```
armiol, line 69: Please mention the implicitly `required` fields. E.g. where they are handled.

armiol (changes requested):
@yevhenii-nadtochii In general, LGTM. Please see my comments.

yevhenii-nadtochii:
@armiol PTAL

java/src/main/kotlin/io/spine/validation/java/ConstraintViolations.kt
```diff
@@ -0,0 +1,97 @@
...
+ * A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+ * OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
```
alexander-yevsyukov, line 26: Please add `@file:JvmName("ConstraintViolations")`.

java/src/main/kotlin/io/spine/validation/java/ConstraintViolations.kt
```diff
@@ -0,0 +1,97 @@
...
+    optionName: String,
+    fieldName: String
+): Expression<TemplateString> {
+    checkPlaceholdersHasValue(template, placeholders.mapKeys { it.key.value }) {
+        "The `($optionName)` option doesn't support the following placeholders: `$it`. " +
+                "The supported placeholders: `${placeholders.keys}`. " +
+                "The declared field: `${fieldName}`."
+    }
+    val placeholderEntries = mapExpression(
+        ClassName(String::class), ClassName(String::class),
```
alexander-yevsyukov, line 90: Why don't we extract the class name of `String` into a constant somewhere? Or, at least, have a variable here.

java/src/main/kotlin/io/spine/validation/java/ConstraintViolations.kt
```diff
@@ -0,0 +1,97 @@
...
+        placeholders.mapKeys { StringLiteral(it.key.toString()) }
+    )
+    return ClassName(TemplateString::class).newBuilder()
```
alexander-yevsyukov, line 93: Similarly to the class name of `String`, we probably would have multiple places that build a `TemplateString` instance. So an extension function or a companion method may be handy.

I'm no insisting at the moment because the code is going to change. But please keep it in mind for following PRs.

java/src/main/kotlin/io/spine/validation/java/OptionGenerator.kt
```diff
@@ -0,0 +1,72 @@
...
+        type: TypeName,
+        parent: Expression<FieldPath>,
+        violations: Expression<MutableList<ConstraintViolation>>
+    ): OptionCode
+}
+
+/**
+ * Java code handling all applications of a specific option within a message.
+ */
+internal class OptionCode(
```
alexander-yevsyukov, line 59: This class deserves to live at the level of the package. This way the reader of the code would pick up this conceptual thing sooner.

model/src/test/kotlin/io/spine/validation/required/RequiredPolicySpec.kt
```diff
@@ -103,3 +98,6 @@ internal class RequiredPolicySpec {
         return error
     }
 }
+
+private fun expected(field: Field) = "The field type `${field.type}` of `${field.qualifiedName}` " +
+        "is not supported by `($REQUIRED)` option."
```
alexander-yevsyukov, line 103: We should probably have the definite article before the name of the option here. Please consult with ChatGPT on this.

alexander-yevsyukov (changes requested):
Please see my comments.

yevhenii-nadtochii:
@alexander-yevsyukov PTAL

armiol (approved):
@yevhenii-nadtochii Looks good, from my side.

alexander-yevsyukov (approved):
LGTM

PR #189, opened 2025-02-11
Generate code for `(pattern)` with a standalone generator

yevhenii-nadtochii:
This PR migrates code generation of `(pattern)` option from the rule-based generator to a standalone one.

The changes are the following:

1. Now we have `PatternGenerator` and `PatternFieldGenerator`. The first class generates code for all marked fields within a message, the latter generates code for a specific field.
2.  I decided to go with shorter name for generators because, anyway, the rule-based ones will be removed, and names ambiguity will be resolved. Thus, `RequiredOptionGenerator` became `RequiredGenerator`. Then, the generators for one field will use `${OptionName}FieldGenerator` name pattern.
3. Moved all rules generation under `RuleGenerator : OptionGenerator`. Though, this class doesn't generate code for a specific option, but for all message rules. Anyway, it is temporary. With this class, we can handle all generators within `JavaValidationRenderer` in a unified manner.
4. I have tried not to pass variables expressions to code generators. They are burden because we should bypass them way deep from a method to method. Instead, they are now `internal` constants of `MessageValidationInjector`.
5. I have split `MessageValidationCode` into data (`MessageValidationCode`) and operations (`MessageValidationInjector`) classes.
6. I've removed a default value for nullable parameter of `constraintViolarion()` creational method. A more popular case is when option generators pass one.


java/src/main/kotlin/io/spine/validation/java/PatternFieldGenerator.kt
```diff
@@ -0,0 +1,226 @@
...
+        fieldType.isSingularString -> {
+            val fieldValue = fieldAccess.getter<String>()
+            val constraint = singularStringConstraint(fieldValue)
+            FieldOptionCode(constraint, pattern.field)
+        }
+
+        fieldType.isRepeatedString -> {
+            val fieldValue = fieldAccess.getter<List<String>>()
+            val validateRepeatedField = mangled("validate$camelFieldName")
+            val validateRepeatedFieldDecl = validateRepeatedField(fieldValue, validateRepeatedField)
```
yevhenii-nadtochii, line 91: As for now, we have to repeat ourselves.

java/src/main/kotlin/io/spine/validation/java/PatternFieldGenerator.kt
```diff
@@ -0,0 +1,226 @@
...
+            val constraint = repeatedStringConstraint(fieldValue, validateRepeatedField)
+            FieldOptionCode(constraint, pattern.field, validateRepeatedFieldDecl)
+        }
+
+        else -> error {
+            "Unsupported field type: `${fieldType.name}`. The `(${PATTERN})` option can be" +
+                    "applied only to singular or repeated string fields."
+        }
```
yevhenii-nadtochii, line 99: I have tried to get rid of this nonsense check because the policy already confirmed the field type. Couldn't find a solution. Information about the confirmed field type can arrive in the event, but Proto enums inevitably have two `service` constants, leading to the same `else` branch in exhaustive `when`.

yevhenii-nadtochii:
@armiol @alexander-yevsyukov please take a preliminary look.

I haven't yet extracted common parts from generators because we have only two of them. 

The PR doesn't build because it requires a single change to ProtoData:

`public sealed class FieldDeclaration<T>(name, code)` should become `open class` instead. This is actually an abstract base, but allowing its instantiation from a raw `code` simplifies the conversion of Poet's `Field` to our `Field` in `RuleGenerator : OptionGeneraot`.

alexander-yevsyukov (changes requested):
Please see my comments.

java/src/main/kotlin/io/spine/validation/java/PatternFieldGenerator.kt
```diff
@@ -0,0 +1,246 @@
...
+            val fieldValues = fieldAccess.getter<List<String>>()
+            val validateRepeatedField = mangled("validate$camelFieldName")
+            val validateRepeatedFieldDecl = validateRepeated(fieldValues, validateRepeatedField)
+            val constraint = repeatedStringConstraint(fieldValues, validateRepeatedField)
+            FieldOptionCode(constraint, listOf(pattern.field), listOf(validateRepeatedFieldDecl))
+        }
+
+        else -> error {
+            "Unsupported field type: `${fieldType.name}`. The `(${PATTERN})` option can be" +
+                    "applied only to singular or repeated string fields."
```
armiol, line 104: In any case, let's add a space between "can be" and "applied".

model/src/main/kotlin/io/spine/validation/core/EventMessageExts.kt
```diff
@@ -0,0 +1,35 @@
...
+
+package io.spine.validation.core
+
+import io.spine.base.EventMessage
+import io.spine.server.event.Just
+
+/**
+ * Converts this [EventMessage] into [Just].
+ */
+public fun <E : EventMessage> E.just(): Just<E> = Just(this)
```
armiol, line 35: Eventually, this should go to Spine Core or Spine Server.

armiol (approved):
@yevhenii-nadtochii with the exception for a couple of comments, LGTM.

Please make sure to address the remaining comments prior to merging.

yevhenii-nadtochii:
@alexander-yevsyukov PTAL

alexander-yevsyukov (approved):
LGTM

PR #190, opened 2025-02-18
Generate code for `(goes)` with a standalone generator

yevhenii-nadtochii:
This PR migrates `(goes)` option to a standalone generator.

- `GoesGenerator` – generates code for all `(goes)` fields within the provided message.
- `GoesFieldGenerator` – generates code for a particular application of `(goes)`.

### Tests for `(goes)`

This is a recently introduced option, which lacked some tests. I have added tests for compilation errors (we have 4 cases here) and tests that verify the created instances of `ConstraintViolation`. 

Now, this option is fully covered.

### Other tests-related changes

1. I've synchronized `SetOnceViolationITest` with a newer `GoesViolationITest` implementation because these test suites are of the same nature. 
2. Moved all `*TestEnv` classes I've met to `test-fixtures` source set. Being inside of `test-fixtures`, they no longer need `given` package.
3. Replaced a special treatment of `Enum` and `Map` in `(goes)/(set_once)` tests with `Builder.set(field, value)` extension.
4. Extensions for `Class<out Message>` were moved to `:model` test fixtures because policy tests also need them. Thus, `:java-tests:validating` now depend on these fixtures from `:model`.
5. I've renamed error message tests to `*ViolationITest` because they actually assert the whole instance of `ConstraintViolation`.

### Changes to `SetOnceJavaConstraints` hierarchy

In `(goes)`, we need to convert field values to `String`. Now we have an extension for this: `FieldType.stringValueOf(value: Expression<*>)`. 

So, I've employed this extension for `(set_once)` generation as well. Unfortunately, it still doesn't work out for `SetOnceStringField` and `SetOnceEnumField` classes because an internal representation of these fields differs from the field type. `String` is stored as `StringOrByteString` value, and `Enum` is stored as `Int`.


yevhenii-nadtochii:

alexander-yevsyukov:
@yevhenii-nadtochii,
 
> ProtoData suggestions

Agreed to all of them.

model/src/main/kotlin/io/spine/validation/FieldIds.kt
```diff
@@ -0,0 +1,37 @@
...
+ */
+
+package io.spine.validation
+
+import io.spine.protodata.ast.Field
+
+/**
+ * Creates a new instance of [FieldId] for this [Field].
+ */
+public fun Field.id(): FieldId = fieldId {
```
alexander-yevsyukov, line 34: This should probably be a part of ProtoData API, but under the name `FieldRef`.

alexander-yevsyukov: And even though this code produces a value, it would read better: `Field.reference` or `Field.ref`.

model/src/main/kotlin/io/spine/validation/PatternPolicy.kt
```diff
@@ -79,16 +77,14 @@ internal class PatternPolicy : Policy<FieldOptionDiscovered>() {
...
-    val type = field.type
-    if (!(type.isSingularString || type.isRepeatedString)) {
-        Compilation.error(file, field.span) {
-            "The field type `${field.type}` of `${field.qualifiedName}` is not supported " +
-                    "by the `($PATTERN)` option. Supported field types: strings and repeated " +
-                    "of strings."
-        }
+private fun checkFieldType(field: Field, file: File) =
+    Compilation.check(field.type.isSupported(), file, field.span) {
+        "The field type `${field.type}` of `${field.qualifiedName}` is not supported " +
```
alexander-yevsyukov, line 82: Let's have the convention for new lines in concatenations: "They start from the space character". It's easier to spot and understand as a continuation. Oftentimes, people miss a space character in concatenated messages. With the starting space it happens less often.

alexander-yevsyukov: Please also add this as a rule in our Code Style Wiki.

model/src/main/kotlin/io/spine/validation/protodata/TypeSystem.kt
```diff
@@ -0,0 +1,39 @@
...
+import io.spine.protodata.ast.TypeName
+import io.spine.protodata.type.TypeSystem
+
+/**
+ * Returns a message type by its name.
+ *
+ * @throws IllegalArgumentException if the requested type is unknown for this [TypeSystem].
+ */
+public fun TypeSystem.message(type: TypeName): MessageType =
+    findMessage(type)?.first ?: throw IllegalArgumentException("Message type not found: `$type`.")
```
alexander-yevsyukov, line 39: Why not just `error()`?

alexander-yevsyukov: `$type` should be `${type.qualifiedName}` or something like that. We don't want to have a Protobuf message printed in the error message.

model/src/test/kotlin/io/spine/validation/AbstractCompilationTest.kt
```diff
@@ -0,0 +1,59 @@
...
+internal abstract class AbstractCompilationTest {
+
+    @TempDir
+    private lateinit var workingDir: Path
+
+    /**
+     * Asserts that the messages represented by the given [descriptor]
+     * fails the compilation process.
+     */
+    fun assertCompilationFails(descriptor: Descriptors.Descriptor): Compilation.Error {
```
alexander-yevsyukov, line 48: Please import `Descriptors.Descriptor`.

Also, this class is a good candidate for ProtoData TestLib.

model/src/test/kotlin/io/spine/validation/AbstractCompilationTest.kt
```diff
@@ -0,0 +1,59 @@
...
+import org.junit.jupiter.api.assertThrows
+import org.junit.jupiter.api.io.TempDir
+
+/**
+ * An abstract base for classes that test compilation errors.
+ */
+internal abstract class AbstractCompilationTest {
+
+    @TempDir
+    private lateinit var workingDir: Path
```
alexander-yevsyukov, line 42: We probably should have this field `protected` for wider variety of descendants.

alexander-yevsyukov (changes requested):
In overall, it looks good. But needs a bit more polishing. Please see my comments.

model/src/testFixtures/proto/spine/validation/goes_bad_target_type.proto
```diff
@@ -0,0 +1,107 @@
+/*
+ * Copyright 2024, TeamDev. All rights reserved.
```
armiol, line 2: 2025.

armiol (changes requested):
@yevhenii-nadtochii in general, looks really good!

Please see my comments.

model/src/main/kotlin/io/spine/validation/PatternPolicy.kt
yevhenii-nadtochii: I've added this convention for [Kotlin](https://github.com/SpineEventEngine/documentation/wiki/Kotlin-code-style#formatting-string-literals) and [Java](https://github.com/SpineEventEngine/documentation/wiki/Java-code-style#formatting-string-literals) styles.

java/src/main/kotlin/io/spine/validation/java/FieldValues.kt
```diff
@@ -0,0 +1,71 @@
...
+            "Cannot convert `$value` expression to `String` expression. " +
+                    "Unsupported field type: `${name}`."
+        )
+    }
+
+@Suppress("UNCHECKED_CAST") // Casting string field's value to `String` is safe.
+private fun Expression<*>.stringifyPrimitive(primitive: PrimitiveType) =
+    when (primitive) {
+        TYPE_STRING -> this as Expression<String>
+        TYPE_BYTES -> stringify()
```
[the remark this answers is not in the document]
yevhenii-nadtochii: I've expanded docs to this function to cover this question. 

>> It may even break some things if the byte array is too big.

There would be `ByteString` in the generated code. It has a safe implementation of `toString()`.

yevhenii-nadtochii:

armiol (approved):

java-tests/validating/src/testFixtures/kotlin/io/spine/test/options/goes/GoesViolationTestEnv.kt
```diff
@@ -0,0 +1,63 @@
...
+        Field("message", Timestamp.getDescriptor().fullName, Timestamps.now()),
+        Field("enum", EnumForGoes.getDescriptor().fullName, EnumForGoes.EFG_ITEM1),
+        Field("string", "string", "some companion text"),
+        Field("bytes", "bytes", ByteString.copyFromUtf8("some companion data")),
+        Field("repeated", "repeated int64", listOf(1L, 2L, 3L)),
+        Field("map", "map<string, int32>", mapOf("key" to 32)),
+    ).map { arguments(named(it.name, "${it.name}_field"), it.type, it.value) }
+}
+
+private class Field(
```
alexander-yevsyukov, line 59: Maybe `data class`?

java-tests/validating/src/testFixtures/kotlin/io/spine/test/options/setonce/SetOnceViolationTestEnv.kt
```diff
@@ -0,0 +1,112 @@
...
+        Primitive("fixed32", "cash_CHF", TWO, EIGHT),
+        Primitive("fixed64", "cash_CNY", TWENTY, SEVENTY),
+        Primitive("sfixed32", "cash_PLN", TWO, EIGHT),
+        Primitive("sfixed64", "cash_NZD", TWENTY, SEVENTY),
+        Primitive("bool", "has_medals", YES, NO),
+        Primitive("bytes", "signature", CERT1, CERT2),
+    ).map { arguments(named(it.type, it.field), it.type, it.value1, it.value2) }
+}
+
+private class Primitive(
```
alexander-yevsyukov, line 107: Again why not `data class` for the sake of clarity?

alexander-yevsyukov (approved):
LGTM, with two minor requests to address before merging.

PR #192, opened 2025-02-26
Generate code for `(distinct)` with a standalone generator

yevhenii-nadtochii:
This PR migrates `(distinct)` option from Rule-based codegen to a standalone generator.

The changes are the following:

1. Implemented `DistinctGenerator` and `DistinctFieldGenerator`.
2. Adjusted `DistinctITest` to test `map` fields and assert `field.duplicates` placeholder value. I think this is the very least of what should be tested. Both primitive and message collections are checked.


yevhenii-nadtochii:

I will extract common traits of generators with a separate branch.

armiol:
@yevhenii-nadtochii Could you please expand the PR description in the section of ProtoData suggestions? Right now it looks a bit unfinished.

yevhenii-nadtochii:
> @yevhenii-nadtochii Could you please expand the PR description in the section of ProtoData suggestions? Right now it looks a bit unfinished.

I've added the description. This PR suggests only one extension.

java-tests/validating/src/testFixtures/proto/spine/test/tools/validate/distinct.proto
```diff
@@ -36,7 +36,7 @@ option java_multiple_files = true;
 
 import "google/protobuf/any.proto";
 
-message ProtoSet {
-
-    repeated google.protobuf.Any element = 1 [(distinct) = true];
+message UniqueCollections {
```
armiol, line 39: Please consider including a more complex use-case in testing. If I recall correctly, we also support the distinct map values (in addition to keys, which are unique by design). So let's test it with `Message`s as well.

The same applies the `repeated` field. We do support `Message`s, right?

armiol (changes requested):
@yevhenii-nadtochii LGTM in general. Please see my question (and a comment on the PR level).

alexander-yevsyukov:
> An extension to look for a field option by name: Field.findOption(name: String): Option?.

Why don't to find it by its type (via descriptor)? And via a generic parameter in an inline function?

Names are fragile usually. I don't think we would be renaming usually, but searching for strings is more time consuming than searching for types.

model/src/main/kotlin/io/spine/validation/OptionNames.kt
```diff
@@ -98,3 +98,8 @@ public const val WHEN: String = "when"
  * The name of the `(goes)` option.
  */
 public const val GOES: String = "goes"
+
+/**
+ * The name of `(if_has_duplicates)` option.
+ */
+public const val IF_HAS_DUPLICATES: String = "if_has_duplicates"
```
alexander-yevsyukov, line 105: Can we get it via descriptors of options instead?

model/src/main/kotlin/io/spine/validation/protodata/FieldExts.kt
```diff
@@ -0,0 +1,36 @@
...
+
+package io.spine.validation.protodata
+
+import io.spine.protodata.ast.Field
+import io.spine.protodata.ast.Option
+
+/**
+ * Looks up a field option by its [name].
+ */
+public fun Field.findOption(name: String): Option? =
```
alexander-yevsyukov, line 35: Again, let's find it via enclosed type instead. Names are fragile.

alexander-yevsyukov (changes requested):
Please see my comments.

model/src/main/kotlin/io/spine/validation/OptionNames.kt
armiol: @yevhenii-nadtochii as we have discussed here vocally just now, let's continue with the string values for now. 

Once we migrate the whole codebase to the current "new" validation approach, we'll be able to get rid of the string literals.

model/src/main/kotlin/io/spine/validation/CompanionErrorMessage.kt
```diff
@@ -0,0 +1,63 @@
...
+ * This method considers two cases:
+ *
+ * 1. If the passed [field] has the option of type [T] applied, the value of its
+ * `error_msg` field is returned (companion options must declare this field).
+ * 2. If the passed [field] does NOT have such an option applied,
+ * the [default error message][DefaultErrorMessage] for [T] is returned.
+ */
+internal inline fun <reified T : Message> resolveErrorMessage(field: Field): String {
+    val descriptor = T::class.descriptor
+    val errorMessageField = descriptor.field(CUSTOM_MESSAGE_FIELD)
```
yevhenii-nadtochii, line 52: Fetches a field descriptor by the field name. 

I haven't found how to get it in a typed manner for `T : Message`. Fetching by `srtring` type looks too fragile, taking into account that in some options, we have a deprecated `string msg_format` field remained.

Another alternative would be to have a similar method declared four times in policies (by the number of boolean options):

```kotlin
private fun resolveErrorMessage(field: Field): String {
    val customMessage = field.findOption<IfMissingOption>()?.errorMsg
    return customMessage ?: IfMissingOption.getDescriptor().defaultMessage
}
```

yevhenii-nadtochii:

java-tests/validating/src/test/kotlin/io/spine/test/options/DistinctITest.kt
```diff
@@ -26,58 +26,184 @@
...
+    `when handling 'list' field` {
+
+        @Test
+        fun `duplicated primitive entries result in a violation`() {
+            val duplicate1 = 123
+            val duplicate2 = 321
+
+            val exception = assertThrows<ValidationException> {
+                uniquePrimitiveCollections {
+                    element.add(duplicate1)
```
alexander-yevsyukov, line 58: How about `element.run { ... }`? I'm not insisting, just asking.

java-tests/validating/src/test/kotlin/io/spine/test/options/DistinctITest.kt
```diff
@@ -26,58 +26,184 @@
...
-            .comparingExpectedFieldsOnly()
-            .containsExactly(expected)
+        @Test
+        fun `duplicated message entries result in a violation`() {
+            val duplicate1 = hoursAndMinutes(2, 20)
+            val duplicate2 = hoursAndMinutes(18, 40)
+
+            val exception = assertThrows<ValidationException> {
+                uniqueMessageCollections {
+                    element.add(duplicate1)
```
alexander-yevsyukov, line 83: Similar question on `element.run { ... }` here. Just to same some eye movements for the sake of readers who struggle not to start wearing glasses.

model/src/testFixtures/kotlin/io/spine/validation/MessageClass.kt
```diff
@@ -26,28 +26,16 @@
...
-/**
- * Returns a Protobuf descriptor for this message [KClass].
- */
-fun KClass<out Message>.getDescriptor() = java.getDescriptor()
-
 /**
  * Creates a new builder for this message [Class].
  */
-fun Class<out Message>.newBuilder() =
+fun Class<out Message>.newBuilder(): Message.Builder =
```
armiol, line 35: We also have [this thing](https://github.com/SpineEventEngine/base/blob/master/src/main/kotlin/io/spine/protobuf/MessageExts.kt#L77), which caches the results for the repeated calls.

If nothing prevents us from using the utility from `base`, let's do it.

java/src/main/kotlin/io/spine/validation/java/DistinctFieldGenerator.kt
```diff
@@ -0,0 +1,148 @@
...
+     * The resulting expression depends on the [fieldType]:
+     *
+     * 1. For `repeated` fields, it is a field value itself.
+     * 2. For `map` fields, it is a collection of map values.
+     */
+    private fun validatedCollection(): Expression<Collection<*>> = when {
+        fieldType.isList -> getter
+        fieldType.isMap -> getter.call("values")
+        else -> error(
+            "Field type `${fieldType.name}` is not supported by `DistinctFieldGenerator`." +
```
alexander-yevsyukov, line 95: Isn't it `The field type...`?

java/src/main/kotlin/io/spine/validation/java/DistinctFieldGenerator.kt
```diff
@@ -0,0 +1,148 @@
...
+     *
+     * The resulting expression depends on the [fieldType]:
+     *
+     * 1. For `repeated` fields, it is a field value itself.
+     * 2. For `map` fields, it is a collection of map values.
+     */
+    private fun validatedCollection(): Expression<Collection<*>> = when {
+        fieldType.isList -> getter
+        fieldType.isMap -> getter.call("values")
+        else -> error(
```
alexander-yevsyukov, line 94: Is it error or compilation error? If it's compilation error the language should not refer to `DistinctFieldGenerator` or other internals. It should talk to the user in terms of types and options he sees.

java-tests/validating/src/test/kotlin/io/spine/test/options/DistinctITest.kt
```diff
@@ -26,58 +26,184 @@
...
-        msg.validate().shouldBeEmpty()
     }
 
     @Test
-    fun `empty list does not result in a violation`() {
-        val msg = assertDoesNotThrow {
-            ProtoSet.newBuilder().build()
+    fun `empty list and map do not result in a violation`() {
+        assertDoesNotThrow {
+            UniquePrimitiveCollections.newBuilder()
```
alexander-yevsyukov, line 203: Why not have Kotlin syntax here? Just curious.

java/src/main/kotlin/io/spine/validation/java/PatternFieldGenerator.kt
```diff
@@ -99,9 +99,9 @@ internal class PatternFieldGenerator(private val view: PatternField) {
         }
 
         else -> error {
-            "Field type `${fieldType.name}` is not supported by `PatternFieldGenerator`. " +
-                    "Please ensure that the supported field types in this generator match those " +
-                    "used by `PatternPolicy` when validating the `PatternFieldDiscovered` event."
+            "Field type `${fieldType.name}` is not supported by `PatternFieldGenerator`." +
```
alexander-yevsyukov, line 102: Same comments here. Isn't it `The field type`? And, isn't it a compilation error?

alexander-yevsyukov (changes requested):
Please see my comments.

model/src/main/kotlin/io/spine/validation/CompanionErrorMessage.kt
```diff
@@ -0,0 +1,63 @@
...
+    val customMessage = field.findOption<T>()?.getField(errorMessageField) as String?
```
armiol, line 53: Honestly, I don't like any of the suggested options — neither hard-coding `error_msg`, nor de-facto switch-casing over the existing option types (and in case of match, returning their `errorMsg` field). They are both bad, but I don't see any better alternatives.

But for sure we could try finding the option first returning the default message if not found. And only then execute `descriptor.field(CUSTOM_MESSAGE_FIELD)` and other tricky stuff.

armiol (changes requested):

java/src/main/kotlin/io/spine/validation/java/DistinctFieldGenerator.kt
yevhenii-nadtochii: This is an error. Field type checks are done by policies.

java/src/main/kotlin/io/spine/validation/java/PatternFieldGenerator.kt
yevhenii-nadtochii: > Is it possible not to get into this branch at all, instead of explaining the error at this level of details?

It'd be great. But I haven't figured out how to make it. 

The policy checks the field type, it can bypass the resolved type to the view. I've tried to do this with per-policy enums, but continued running into non-exhaustive `when` because Protobuf enums inevitably have some service constants.

model/src/testFixtures/kotlin/io/spine/validation/MessageClass.kt
yevhenii-nadtochii: I can move this overload to `base`, so it redirects to `builderFor()`.

It just reads more familiar in tests:

image

Compared to:

image

alexander-yevsyukov (approved):
LGTM

yevhenii-nadtochii:
@armiol PTAL

armiol (approved):

PR #194, opened 2025-03-04
Generate code for `(validate)` with a standalone generator

yevhenii-nadtochii:
This PR migrates code generation for `(validate)` option from Rules to a standalone `ValidateGenerator`.

`ValidateConstraintTest` was replaced with `ValidateITest`. It covered a limited number of test cases, so I decided to implement a new one. This is a behavioral test for `(validate)`. 

### API changes in `:java-runtime`

1. `ValidatableMessage` now has both `validate()` and `validate(FieldPath parentPath, TypeName parentName)` methods. The first one has a default implementation that redirects to the second method. We need this method to be able to invoke validation upon a validatable message in a way to preserve information about the parent path and type name. These are used to assemble an instance of `ConstraintViolation` in a way to conform with its docs.
2. I've clarified docs to `ConstraintViolation.type_name`. When in-depth validation takes place, it should contain the name of the message that initiated the validation, not the message containing the invalid field. Below, we have a `field_path`. Its root field should always belong to `type_name`. All option generators have been updated to take this into account.

### Other changes

1. Removed illegal applications of `(validate)` from test fixtures. They are compilation errors now.
2. Updated tests that should have been asserting a nested path from the very beginning.
3. Removed `validating_builder_test.proto` – had no usages.
4. Removed dependency on `:java-runtime` from modules that use dogfooded version of `mc-java`. They should use their own version of `java-runtime`.

### Left TODOs

1. `ValidationCodeInjector` should not render the default implementation of `validate()` from `ValidatableMessage`. It does this temporarily to keep binary compatibility until ProtoData get the latest version of `validation-runtime` for its uber jar. See #198.
2. `EmptyFieldCheck` can be implemented more efficiently, and without `JavaValueConverter` (which requires `TypeSystem`). See #199.


yevhenii-nadtochii:
@armiol @alexander-yevsyukov PTAL and let's discuss.

1. Do you have any assumption regarding the bug?
2. About `ValidatableMessage .validate()`. Should we have `TypeName` (to avoid dealing with string) in `base` OR extend `FieldPath` to have `root_message_name` property? So that the field path is fully-navigatable?
3. Confirm the option behavior, so that we are OK upon how it behaves with `Any` and default instances.

armiol:
@yevhenii-nadtochii as discussed, we probably know how to deal with the bug. Please act according to the plan, and let us know once you are done.

java/src/main/kotlin/io/spine/validation/java/generate/ValidationCodeInjector.kt
```diff
@@ -129,49 +129,27 @@ private fun MessagePsiClass.implementValidatableMessage() {
...
  *
- * This is a `public` implementation of [ValidatableMessage.validate] with
- * [Override] annotation. The actual constraints are contained in its
- * private overloading, that accepts the field path.
+ * This method implements the logic for verifying that the message’s constraints are met.
+ * It takes the parent path and name as arguments to preserve this information for cases
+ * when in-depth validation takes place. This data is used to construct constraint violations.
  *
- *  @see declarePrivateValidateMethod
+ * In typical (top-level) validations, the [ValidatableMessage.validate] method is called,
```
armiol, line 138: I don't think "validation" exists in plural. Can we rephrase this as "In typical use cases of validating the top-level messages ..."?

java/src/main/kotlin/io/spine/validation/java/generate/option/ValidateFieldGenerator.kt
```diff
@@ -0,0 +1,167 @@
...
+     *
+     * The default instances of [com.google.protobuf.Message] are considered valid
+     * for singular fields even if the checked message has one or more required fields.
+     * In this case, it is impossible to determine whether the field itself is set
+     * with an invalid message instance or just empty.
+     *
+     * The above statement doesn't apply to repeated and map fields. Within collections,
+     * default instances are considered invalid.
+     *
+     * Details regarding the generated code:
```
armiol, line 126: Honestly, both points are really not easy to comprehend, let alone update after any modifications are made to the code in the future.

I would remove this block altogether.

armiol (changes requested):
@yevhenii-nadtochii excluding the code related to the "bugged" test (which is in progress), LGTM.

See a couple of comments, though.

java-runtime/src/main/java/io/spine/validate/ValidatableMessage.java
```diff
@@ -44,5 +45,25 @@ public interface ValidatableMessage extends Message {
...
+     * to the original field that initiated in-depth validation.
+     *
+     * @param parentFieldPath
+     *         the path to the parent field that initiated in-depth validation
+     * @param parentTypeName
+     *         the name of the parent type that initiated in-depth validation
+     *
+     * @return an error or {@link Optional#empty()} if no violations found
+     */
+    Optional<ValidationError> validate(FieldPath parentFieldPath, String parentTypeName);
```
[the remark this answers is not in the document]
alexander-yevsyukov: We discussed with @armiol vocally that the second parameter could have the type `io.spine.type.TypeName` which comes from Base. You probably discuss it too. We have the dependency on Base anyway, and having `TypeName` from there is better than just `String`.

java-runtime/src/main/proto/spine/validate/validation_error.proto
```diff
@@ -62,6 +62,28 @@ message ConstraintViolation {
...
+    // If the `(validate)` option is applied and a nested field contains an invalid value,
+    // this property holds the name of the root message, which triggered the validation.
+    //
+    // Example:
+    //
+    // ```
+    // message Student {
+    //     Contacts contacts = 1 [(validate) = true];
+    // }
+    // message Contacts {
```
alexander-yevsyukov, line 75: Please have empty lines between the examples. Yes, it would make the block bigger, but a code of an example should follow the normal style people write the code in the language of the example. We do have empty lines between type declarations in the production code.

alexander-yevsyukov (approved):
LGTM, with comments to consider.

java-runtime/src/main/java/io/spine/validate/ValidatableMessage.java
```diff
@@ -44,5 +47,24 @@ public interface ValidatableMessage extends Message {
...
+     * Validates this message according to the rules in the Protobuf definition.
+     *
+     * <p>Use this overload when validating a message as part of another message's validation.
+     * In this case, any constraint violations reported by this method will include the path
+     * to the original field that initiated in-depth validation.
+     *
+     * @param parentPath
+     *         the path to the parent field that initiated in-depth validation
+     * @param parentName
+     *         the name of the parent type that initiated in-depth validation
```
armiol, line 65: ... and let's explain what `null` means.

java/src/main/kotlin/io/spine/validation/java/generate/option/RequiredFieldGenerator.kt
```diff
@@ -72,26 +76,32 @@ internal class RequiredFieldGenerator(
             """
             if (${field.hasDefaultValue()}) {
                 var fieldPath = ${parentPath.resolve(field.name)};
-                var violation = ${violation(ReadVar("fieldPath"))};
+                var typeName =  $parentName != null ? $parentName : $TypeNameClass.of(this);
```
armiol, line 79: As a side comment.

This thing:
```
 ... $parentName != null ? $parentName : $TypeNameClass.of(this)
```
is now a frequent guest throughout this PR.  Just use a search by `$parentName != null ?`.

Please consider extracting it somehow.

armiol (changes requested):
@yevhenii-nadtochii please see my comments. Otherwise, LGTM.

yevhenii-nadtochii:
@armiol PTAL

armiol (approved):

PR #195, opened 2025-03-04
Fix escaping slash char

alexander-yevsyukov:
@yevhenii-nadtochii, I've created this PR to show the escaping issue we have with the `/` character in ProtoData. Please see `patterns.proto` with two new entries for the docs on what goes wrong.

I've experienced the issue when trying to build new ProtoData with new Validation. Please pick up this PR and fix the escaping.

If possible, please avoid using `org.apache.commons.lang` or other transitive utilities that may come via PSI to us. If you need escaping please use it from Guava on which we standardise. Or, which is more preferable, please find a Kotlin utility which does the same.

We should generally avoid having direct a dependency on Apache Commons or similar projects from Apache.

### DIFF
--- a/java/src/main/kotlin/io/spine/validation/java/generate/option/PatternFieldGenerator.kt
+++ b/java/src/main/kotlin/io/spine/validation/java/generate/option/PatternFieldGenerator.kt
@@ -70,7 +70,6 @@ import io.spine.validation.java.generate.mangled
 import io.spine.validation.java.violation.templateString
 import java.util.regex.Matcher
 import java.util.regex.Pattern
-import org.apache.commons.lang.StringEscapeUtils.escapeJava
 
 /**
  * Stores the generated Java field that contains the compiled [Pattern]
@@ -184,7 +183,7 @@ internal class PatternFieldGenerator(private val view: PatternField) : FieldOpti
     private fun compilePattern(): CompiledPattern {
         val modifiers = view.modifier
         val compilationArgs = listOf(
-            StringLiteral(escapeJava(view.pattern)),
+            StringLiteral(reEscapeProtobuf(view.pattern)),
             Literal(modifiers.asFlagsMask())
         )
         val field = FieldDeclaration<Pattern>(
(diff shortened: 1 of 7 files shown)


yevhenii-nadtochii:
@armiol @alexander-yevsyukov please take a look on the suggested fix.

Touching the complex topic, [specs](https://github.com/SpineEventEngine/base/blob/e9cc400d8b95fc2e4981e679f26c94bc7c5b72a0/src/main/proto/spine/options.proto#L806) to `PatternOption` are not really descriptive about what can go to `regex` field. There are number of Regex dialects with their own relatively small differences in syntax and escaping rules (on the language compiler level and, optionally, on their Regex engine level).

```
    // The regular expression to match.
    string regex = 1;
```

Now, the option feels a bit Java-centric, not language-agnostic. And this is not reflected in docs. This field is more like `string java_regex = 1;` because I am not sure we can accept sort of universal Regex with a predicted behavior for Java/JS/Dart/etc.

Let's take a look on how it goes with the literal caused the initial error:

1. The original literal as defined in a `.proto` file: `"\\/"`.
2. After Protobuf compiler: `"\\/" -> "\/"`. 
3. After `PatternFieldGenerator` (aims to cancel Protobuf actions): `"\/" -> "\\/"`.
4. After Java compiler: `"\\/" -> "\/"`.
5. After Regex compiler: `"\/" -> "/"`.

Regarding (5): Java Regex treats `\` as en escape for `/`, which indeed does not even have to be escaped. So, `[^\\/]` is the same as `[^/]`.

java/src/main/kotlin/io/spine/validation/java/generate/option/PatternFieldGenerator.kt
```diff
@@ -252,3 +251,52 @@ private fun PatternOption.Modifier.asFlagsMask(): Int {
...
+)
+
+/**
+ * Undoes [ASCII escaping][ProtobufEscape] that Protobuf compiler performs for string literals,
+ * allowing the passed [value] to be used as a string literal in Java code.
+ *
+ * Consider the following example:
+ *
+ * 1. A string literal is defined in a proto file as the following: `[^\\/]`.
+ * 2. Runtime string representation would be the following: `[^\/]`.
```
armiol, line 289: Please add more steps to this example so that it is absolutely clear what happens if we don't escape the code on our side again.

Right now, this example says nothing about our code generation, and the state of things in the generated code.

armiol (changes requested):
@yevhenii-nadtochii please see my comment.

It's rather challenging to follow the un-/re-escaping ideas without understanding the full picture. In addition to my comment, I also wonder how we suggest that framework users  do not forget to call this re-escaper.

armiol:
@yevhenii-nadtochii 
> ... `[^/]`

Just in case, `[^/]` may be an invalid regular expression, if `/` is used as the delimiter. Use something like https://regex101.com/ to play with it.

yevhenii-nadtochii:
Addressed in #197.

PR #196, opened 2025-03-06
Bump Gradle, Kotlin and Java

yevhenii-nadtochii:
This PR takes the latest `config`, bumping the version of Gradle, Kotlin and Java.

Adjustments are the following:

1. In `module.gradle.kts`: we can't longer disable incremental compilation using DSL block. Only a Gradle property can be set. See [Kotlin | Incremental Compilation](https://kotlinlang.org/docs/gradle-compilation-and-caches.html#incremental-compilation) for details. I've added a property to `gradle.properties` file. Anyway, we had it disabled for every module.
2. Specified an explicit dependency `compileKotlin.dependsOn(kspKotlin)` for `:java-tests:extensions` module as required by Gradle.
3. Removed unnecessary `public` visibility as reported by the Kotlin compiler.
4. Clarified `T` type parameter as reported by the Kotlin compiler.

The library version was bumped to `2.0.0-SNAPSHOT.300`.


yevhenii-nadtochii:

PR #200, opened 2025-03-13
Generate code for `(range)` with a standalone generator

yevhenii-nadtochii:
This PR migrates the `(range)` option to the code generation with a standalone generator. 

Now we have `RangeGenerator` and `RangeFieldGenerator`, which are responsible for Java code of the option.

### Unsigned primitive types

The model correctly parses and handles unsigned values (`UInt` and `ULong` simplified things a lot). 

In Java codegen, we decided to support unsgined primitives in a way Protobuf does it: unsigned numbers are just stored in signed primitives. All operations upon them should be performed with dedicated static methods that treat signed numbers as if they were unsigned. So, we employed these methods for the comparisons.

Additionally, `RangeFieldGenerator` prints a warning whenever it generates Java code for unsigned primitives. 

Though, now these warnings pollute the build log a bit.


java-tests/validating/src/testFixtures/proto/spine/test/tools/validate/range_field_extrema.proto
```diff
@@ -0,0 +1,55 @@
...
+
+option (type_url_prefix) = "type.spine.io";
+option java_package = "io.spine.validation";
+option java_outer_classname = "RangeFieldExtremaProto";
+option java_multiple_files = true;
+
+// A message that tests the `(range)` option handles boundary values for the supported
+// field types. These values represent the minimum and maximum values a specific field
+// type can be assigned.
+message RangeFieldExtrema {
```
yevhenii-nadtochii, line 42: The word `boundary` is already taken to represent range values. I've used `extrema` (latin plural) from math.

model/src/main/kotlin/io/spine/validation/PatternPolicy.kt
```diff
@@ -93,12 +92,12 @@ private fun FieldType.isSupported(): Boolean = isSingularString || isRepeatedStr
  * The property is `public` because the option generator also uses it.
  */
 public val FieldType.isRepeatedString: Boolean
-    get() = isList && list.primitive == TYPE_STRING
```
yevhenii-nadtochii, line 96: We don't have to check it because otherwise `list.primitive` would have a default value, which is not `TYPE_STRING`. The same is above.

proto/context/src/main/proto/spine/validation/events.proto
```diff
@@ -71,11 +72,11 @@ message RequiredFieldDiscovered {
 
     protodata.FieldRef id = 1;
 
-    // The error message template.
-    string error_message = 2;
-
     // The field in which the option was discovered.
-    protodata.Field subject = 3;
+    protodata.Field subject = 2;
```
yevhenii-nadtochii, line 76: I've re-ordered fields, so that the most common `id, subject, error_message` go first, and the option-specific fields go last.

yevhenii-nadtochii:

I will bring the latest ProtoData with a separate PR because it requires more changes.

model/src/main/kotlin/io/spine/validation/range/NumericBound.kt
```diff
@@ -0,0 +1,166 @@
...
+/**
+ * One to one Kotlin representation of [ProtoNumericBound].
+ *
+ * We would like to have a Kotlin counterpart because of [UInt] and [ULong] types.
+ * It eases the implementation of parsing and comparisons for such bounds.
+ * Otherwise, we would have to care for special cases with unsigned types (which are just `int`
+ * and `long` in Protobuf) when parsing and comparing such bounds. With this class,
+ * Kotlin unsigned classes do it for us.
+ */
+internal data class NumericBound(
```
armiol, line 58: I don't get the idea of having the exact same name for this data class as for the original `NumericBound`.

Imagine I write a code in this module, and the IDE prompts me to import either
* `io.spine.validation.NumericBound`, which is the Proto message, or
* `io.spine.validation.range.NumericBound`, which is this data class.

Let's try to reduce the amount of mental energy required from the developer.

model/src/main/kotlin/io/spine/validation/range/NumericBound.kt
```diff
@@ -0,0 +1,166 @@
...
+    val value: Any, // Cannot use `Number` because Kotlin's `UInt` and `ULong` are not numbers.
+    val inclusive: Boolean
+) : Comparable<NumericBound> {
```
armiol, line 61: Isn't it possible to make the Proto `NumericBound` implement some interface, which would extend `Comparable`, with the default implementation of `compareTo`?

model/src/main/kotlin/io/spine/validation/range/RangeContext.kt
```diff
@@ -0,0 +1,41 @@
...
+ */
+
+package io.spine.validation.range
+
+import io.spine.protodata.ast.Field
+import io.spine.protodata.ast.File
+import io.spine.protodata.ast.PrimitiveType
+
+/**
+ * Data required to report a compilation error for the `(range)` option.
```
armiol, line 34: This part of explanation is needed, for sure. But it reads as the second paragraph.

The first one should be something like 
```
The context of evaluating the `(range)` option ...
```

model/src/main/kotlin/io/spine/validation/range/RangePolicy.kt
```diff
@@ -0,0 +1,188 @@
...
+
+/**
+ * Controls whether a field should be validated with the `(range)` option.
+ *
+ * Whenever a field marked with `(range)` option is discovered, emits
+ * [RangeFieldDiscovered] event if the following conditions are met:
+ *
+ * 1. The field type is supported by the option.
+ * 2. The passed regex has the correct syntax and is compatible with the target field type.
+ *    2.1. Either `..` or ` .. ` is used as a range delimiter.
```
armiol, line 74: Do we really need these double-numbered points? It would be helpful in a document with many points, which reference each other. Here it is just weird.

armiol (changes requested):
@yevhenii-nadtochii LGTM in general. Please see some comments.

java/src/main/kotlin/io/spine/validation/java/generate/option/RangeFieldGenerator.kt
```diff
@@ -0,0 +1,214 @@
...
+     */
+    private fun doesNotBelongToRange(value: Expression<Number>): Expression<Boolean>  {
+        val valueCase = lower.valueCase // This case is the same for both `lower` and `upper`.
+        val lowerLiteral = lower.asLiteral()
+        val lowerOperator = if (lower.inclusive) "<" else "<="
+        val upperLiteral = upper.asLiteral()
+        val upperOperator = if (upper.inclusive) ">" else ">="
+        if (valueCase in listOf(UINT32_VALUE, UINT64_VALUE)) {
+            Compilation.warning(view.file, field.span) {
+                "Unsigned integer types are not natively supported in Java." +
```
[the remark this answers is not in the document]
yevhenii-nadtochii: Neither Protobuf nor Java describes this moment somewhere. I have added a link to a table with Proto->Java mapping, and an article from Baeldnug describing unsigned API.

model/src/main/kotlin/io/spine/validation/range/NumericBound.kt
yevhenii-nadtochii: It is possible to create a `Mixin`, but with a Kotlin class we don't do the work twice.

Now we parse `string` to a Kotlin type, compare it with them and transform to Proto.

With `Mixin`, we parse `string` to a Kotlin type (to know it parses well), set the parsed value to a Proto message (throwing away already created Kotlin instance). And when comparing, create a Kotlin type from Proto for comparisons.

yevhenii-nadtochii:
@armiol PTAL

armiol (approved):

java/src/main/kotlin/io/spine/validation/java/generate/option/RangeFieldGenerator.kt
```diff
@@ -0,0 +1,224 @@
...
+     * Outside of these methods, these primitives remain just signed types.
+     */
+    private fun doesNotBelongToRange(value: Expression<Number>): Expression<Boolean>  {
+        val valueCase = lower.valueCase // This case is the same for both `lower` and `upper`.
+        val lowerLiteral = lower.asLiteral()
+        val lowerOperator = if (lower.inclusive) "<" else "<="
+        val upperLiteral = upper.asLiteral()
+        val upperOperator = if (upper.inclusive) ">" else ">="
+        if (valueCase in listOf(UINT32_VALUE, UINT64_VALUE)) {
+            Compilation.warning(view.file, field.span) {
```
alexander-yevsyukov, line 148: @yevhenii-nadtochii, 

We can probably store a number of times we issue this warning and put the full text only once. On second and after occasions we just print a short text referring to the same issue having some text sounding like "Already reported".

Gradle does a similar thing for repeated errors.

This way the console log won't be "polluted".

Issue #201, opened 2025-03-23
Check for `null` argument in the generated `validate(FieldPath, TypeName)` methods


alexander-yevsyukov:
We need to update our code generation to add null checks for arguments of the generated methods.

### Problem description

I've tried to build CoreJava with the latest ProtoData and McJava and stumbled on the test failure in `FilterMixinTest`. The failed test is `passNullToleranceCheck()` which fails because `NullPointerTester().testAllPublicInstanceMethods(filter)` fails.

It fails because we do not throw `NullPointerException` on `null` value for the `parentPath` parameter which has the `FieldPath` value.

Here's how the generated method of the `Filter` class starts:

```java
@Override
public java.util.Optional<io.spine.validate.ValidationError> validate(io.spine.base.FieldPath parentPath, io.spine.type.TypeName parentName) {
    var violations = new java.util.ArrayList<io.spine.validate.ConstraintViolation>();
...
}
```

We need to have `null` checks for non-nullable parameters as we do in the handwritten Java code.

### The issue with `@Nullable` parameter annotation

The generated code of `validate()` implements the method from the `ValidatableMessage` interface. The method declaration looks like this:

```java
Optional<ValidationError> validate(FieldPath parentPath, @Nullable TypeName parentName);
```

Please note that the generated method of the Filter does not have the `@Nullable` annotation of the `parentName` parameter.
We need to be consistent and either remove the annotation from the interface, or, have it in the generated code and treat the paramter accordingly.

### Implementation preferences

Let's check for `null`s using standard Java API instead of using `Preconditions` from Guava. We need to reduce our dependencies on Guava as we migrate towards Kotlin.

So the suggested implementation is:

```java
Objects.requireNonNull(parentPath, "parentPath");
```

alexander-yevsyukov:
@armiol, FYI.

PR #203, opened 2025-03-25
Propagate `@Nullable` annotation to the generated code and add a precondition check

yevhenii-nadtochii:
This PR adds `@Nullable` annotation to `TypeName` parameter in the generated implementations of `ValidatableMessage.validate(FieldPath, TypeName)` method.

Also, the generated implementations of the above method now check for nullability of `FieldPath` using `Objects.requireNonNull()`.

An example of the generated code:

```java
 @Override
    public java.util.Optional<io.spine.validate.ValidationError> validate(io.spine.base.FieldPath parentPath, io.spine.type.@org.checkerframework.checker.nullness.qual.Nullable TypeName parentName) {
        java.util.Objects.requireNonNull(parentPath);
        // ...
```

Addresses #201.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -92,6 +92,7 @@ allprojects {
                 Roaster.api,
                 Roaster.jdt,
                 ProtoData.api,
+                ProtoData.params,
                 ProtoData.pluginLib,
                 ProtoData.backend,
                 ProtoData.java,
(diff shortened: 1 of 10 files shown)


yevhenii-nadtochii:

armiol (approved):

PR #204, opened 2025-03-27
Generate code for `(min)` and `(max)` using standalone generators

yevhenii-nadtochii:
This PR migrates the `(min)` and `(max)` options to the code generation with standalone generators. 

1. In `:model`, all policies and views related to numeric bounds were placed under the `bound` package. 
2. In `:java`, the generators of min, max and range now have the common parent – `BoundedFieldGenerator`.
3. `NumericBound.inclusive` became `NumericBound.exclusive` because both `MinOption` and `MaxOption` have `exclusive` flag. It makes no difference for the `(range)` option, but brings more consistency with min and max.
4. Error messages that parsed `(range)` bounds became a bit more abstract now, so that this code could also be used to parse standalone min and max bounds. I have removed the word `range` from these literals, and the full `range` itself is no longer printed for bounds-related errors. Only the wrong bound itself. It doesn't apply to errors reported directly from the `RangePolicy`. Only to those, reported during the parsing of `KotlinNumericBound`.

### Removed code

These rules-related files are no longer needed:

1. `NumberRules`.
2. `QueryingExts`.
3. `RangeNotation`.
4. `Numbers`.

### Other changes

Bumped JVM heap size for Gradle to 4gb.


java-tests/runtime/src/test/kotlin/io/spine/validate/NumberRangeSpec.kt
```diff
@@ -106,22 +103,8 @@ internal class NumberRangeSpec : ValidationOfConstraintTest() {
...
     fun `provide one valid violation if number is greater than max`() {
         maxNumberTest(GREATER_THAN_MAX, inclusive = true, valid = false)
-        assertSingleViolation(
-            GREATER_MAX_MSG,
-            VALUE
-        )
-    }
-
-    @Test
-    fun `not allow fraction boundaries for integer fields`() {
```
yevhenii-nadtochii, line 116: This is a compilation error now.

yevhenii-nadtochii:

model/src/main/kotlin/io/spine/validation/RangeNotation.kt
```diff
@@ -1,110 +0,0 @@
...
-    val max: Value
-) {
-
-    companion object {
-
-        /**
-         * The regular expression for parsing number ranges.
-         *
-         * Defines four groups:
-         *  1. The opening bracket (a `[` or a `(`).
```
armiol, line 69: Is there a place to which the helpful set of the examples in this documentation piece migrated?

armiol (approved):
@yevhenii-nadtochii LGTM, but:

1. Please see my question. It would be helpful to migrate the pieces of documentation which are still valid, rather than deleting them with the files.
2. Please make sure that `bound` package is in place not only for Kotlin and Java sources, but also for the matching Proto definitions (i.e. folder structure, Proto package, destination package in Java). It is difficult to verify just by viewing this PR.

Other than that, looks nice.

java/src/main/kotlin/io/spine/validation/java/generate/option/BoundedFieldGenerator.kt
```diff
@@ -0,0 +1,210 @@
...
+            else -> error(
+                "Unexpected field type `$valueCase` when converting range bounds to Java literal." +
+                        " Make sure the policy, which verified `${view::class.simpleName}`," +
+                        " correctly filtered out unsupported field types."
+            )
+        }
+}
+
+private fun unsignedIntegerWarning(file: File, span: Span) =
+    Compilation.warning(file, span) {
```
alexander-yevsyukov, line 196: Reiterating the comment which I [made for already closed PR](https://github.com/SpineEventEngine/validation/pull/200/files#r2012570908):

> [!NOTE]
> We can probably store a number of times we issue this warning and put the full text only once. On second and after occasions we just print a short text referring to the same issue having some text sounding like "Already reported".
>
> Gradle does a similar thing for repeated errors.
>
>  This way the console log won't be "polluted".
>

alexander-yevsyukov: In other words, let's show this long warning text only once for the first field with the unsigned bounds set for a signed type. Then, have a shorter text that mentions the field name, its type and the fact that the full warning was reported before.

@armiol, FYI.

alexander-yevsyukov: This comment applies for the `(range)` option too. We can probably combine the warnings or have their "branches" depending of the option kind.

alexander-yevsyukov (commented):
Please have a look at my comments. I'm not insisting that the commented matter should block this PR. 

We can discuss it and put it off until a later PR, when we have proper support for long and short versions of warnings in ProtoData.

java/src/main/kotlin/io/spine/validation/java/generate/option/BoundedFieldGenerator.kt
alexander-yevsyukov: We can probably have qualifiers for warnings similarly to those that ErrorProne and detekt have. Some titles with `CamelNotation`. 

This will allow:

1. Referencing the full version of text from shorter ones.
2. Suppressing them during the compilation process via ProtoData options.

model/src/main/kotlin/io/spine/validation/RangeNotation.kt
yevhenii-nadtochii: I've moved them to the policy docs.

PR #206, opened 2025-03-28
Make ID policies emit `RequiredFieldDiscovored` event

yevhenii-nadtochii:
This PR updates `RequiredIdOptionPolicy` and `RequiredIdPatternPolicy` to emit `RequiredFieldDiscovored` instead of a simple rule. These policies treat the first fields of signal messages and entity estates as implicitly required.

Note, that If the ID field has the `(required)` option applied explicitly, this is still the responsibility of `RequiredPolicy` to handle it. `RequiredIdPolicy` will ignore it. 

The `RequiredRule` object and `fun Rule.toEvent(type: TypeName)` have been removed as they are no longer needed.


yevhenii-nadtochii:

model/src/main/kotlin/io/spine/validation/required/RequiredIdOptionPolicy.kt
```diff
@@ -33,13 +33,15 @@ import io.spine.protodata.ast.firstField
...
 
 /**
  * A policy that marks ID fields in entity state messages as required.
  *
- * The entity state messages are discovered via the options, specified in [ValidationConfig].
- * If ProtoData runs with no config, this policy never produces any validation rules.
+ * The entity state messages are discovered via the options, specified in
+ * [ValidationConfig][io.spine.validation.ValidationConfig].
+ *
+ * The policy does nothing if ProtoData runs without the config.
```
armiol, line 44: I don't understand what you mean.

If ProtoData runs without a config, many policies will do nothing. Let's just kill this sentence.

model/src/main/kotlin/io/spine/validation/required/RequiredIdPatternPolicy.kt
```diff
@@ -37,13 +36,16 @@ import io.spine.server.event.NoReaction
...
- * A policy that marks ID fields in entity state messages and signal messages as required.
+ * A policy that marks ID fields in entity state messages and signal
+ * messages as required.
  *
- * The messages are discovered via the file patterns, specified in [ValidationConfig].
- * If ProtoData runs with no config, this policy never produces any validation rules.
+ * The messages are discovered via the file patterns, specified in
+ * [ValidationConfig][io.spine.validation.ValidationConfig].
+ *
+ * The policy does nothing if ProtoData runs without the config.
```
armiol, line 48: Same as above.

armiol (changes requested):

Both of your suggestions sound good to me.

model/src/main/kotlin/io/spine/validation/OptionExts.kt
```diff
@@ -31,18 +31,17 @@ package io.spine.validation
 import com.google.protobuf.BoolValue
 import com.google.protobuf.GeneratedMessage.GeneratedExtension
 import io.spine.protobuf.unpack
+import io.spine.protodata.ast.Field
 import io.spine.protodata.ast.Option
 
 /**
- * Checks if this option represents the given generated option.
+ * Finds the option with the given generated type applied to this [Field].
```
alexander-yevsyukov, line 38: Maybe it's "generated extension type"?  I mean all the types generated out of Protobuf definitions are generated. Here the parameter has a precise type.

model/src/main/kotlin/io/spine/validation/required/RequiredIdOptionPolicy.kt
```diff
@@ -33,13 +33,15 @@ import io.spine.protodata.ast.firstField
...
 import io.spine.server.tuple.EitherOf2
-import io.spine.validation.event.RuleAdded
+import io.spine.validation.event.RequiredFieldDiscovered
 
 /**
  * A policy that marks ID fields in entity state messages as required.
  *
- * The entity state messages are discovered via the options, specified in [ValidationConfig].
- * If ProtoData runs with no config, this policy never produces any validation rules.
+ * The entity state messages are discovered via the options, specified in
```
alexander-yevsyukov, line 41: Can we name the part of `ValidationConfig` which is responsible for this?

model/src/main/kotlin/io/spine/validation/required/RequiredIdPatternPolicy.kt
```diff
@@ -37,13 +36,16 @@ import io.spine.server.event.NoReaction
...
+import io.spine.validation.event.RequiredFieldDiscovered
 
 /**
- * A policy that marks ID fields in entity state messages and signal messages as required.
+ * A policy that marks ID fields in entity state messages and signal
+ * messages as required.
  *
- * The messages are discovered via the file patterns, specified in [ValidationConfig].
- * If ProtoData runs with no config, this policy never produces any validation rules.
+ * The messages are discovered via the file patterns, specified in
```
alexander-yevsyukov, line 45: Which part of the `ValidationConfig` define the patterns? Let's direct the user so that we have less support calls.

alexander-yevsyukov (changes requested):
Looks good, but some documentation improvements would be nice to have before merging.

yevhenii-nadtochii:

proto/context/src/main/proto/spine/validation/required/events.proto
```diff
@@ -0,0 +1,59 @@
...
+import "spine/protodata/ast.proto";
+
+// A required field has been discovered.
+//
+// The event emitted in the following cases:
+//
+// 1. A field with `(required) = true` option is discovered.
+// 2. The ID field of a signal message or entity state is identified as required.
+//
+// The event confirms that the necessary checks have been passed, ensuring that
```
armiol, line 47:
```
The event emission happens iff the necessary checks have been passed, ...
```

(See [iff](https://www.dictionary.com/browse/iff) definition, just in case.)

armiol (approved):
@yevhenii-nadtochii LGTM with a comment related to the documentation.

proto/context/src/main/proto/spine/validation/required/events.proto
```diff
@@ -0,0 +1,59 @@
...
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+syntax = "proto3";
+
+package spine.validation;
```
alexander-yevsyukov, line 29: The package is still `spine.validation`, and I don't think we even need a subdirectory for just two types.

Generated code and Kotlin code is a different story. We need packages there so that we can see a bigger picture.
Contrary, on the level of Protobuf, separating just two types makes the whole picture harder to understand. Please revert.

alexander-yevsyukov (changes requested):
Please see my comments.

yevhenii-nadtochii:
@alexander-yevsyukov PTAL

alexander-yevsyukov (approved):
LGTM

PR #207, opened 2025-04-01
Report a compilation error when companion options used alone

yevhenii-nadtochii:
This PR creates policies for `(if_missing)`, `(if_has_duplicates)` and `(if_set_again)` options. These policies report a compilation error if a companion option is applied without its primary boolean counterpart.

Also, policies and views of `(required)`, `(distinct)`, `(validate)`, `(set_once)` are under the same file per each option now. The same is with their tests.

Related changes:

1. Removed no longer valid test cases.
2. Moved the field type check from `SetOnceRenderer` to the corresponding policy.
3. Removed an abstract `ValidationPolicy`. It doesn't have inheritors anymore. 
4. Renamed `ValidationPluginPart` to `WithValidationSettings`. This name fits better for the class. Every policy or view is part of the validation plugin, but they all don't extend this type just because they are part of the plugin. Only policies that handle implicitly required fields use the settings, and extend from this type.
5. Removed `BoolFieldOptionView` and its repository. Now the policies are responsible for that logic.

Additional changes:

1. Added `Supported filed types:` section to the error message of `(goes)`.
2. Bumped the JVM metaspace because the daemon was running out of it.
6. Specified the JVM heap size in MB so that the `org.gradle.jvmargs` property value is the same as in `core-java`.


yevhenii-nadtochii:

model/src/main/kotlin/io/spine/validation/WithValidationSettings.kt
```diff
@@ -33,7 +33,7 @@ import io.spine.protodata.settings.defaultConsumerId
  * An interface common for components of [ValidationPlugin] that load settings
  * passed as a file containing [ValidationConfig] instance written using the plugin class name.
  */
-public interface ValidationPluginPart: LoadsSettings {
+public interface WithValidationSettings: LoadsSettings {
```
armiol, line 36: P.4 from the PR description, namely

> 4. Renamed `ValidationPluginPart` to `WithValidationSettings`. Only implicit required policies use this type.

... hardly explains the reason.

Please elaborate.

armiol (changes requested):
@yevhenii-nadtochii I like the "compacted" layout of the classes more than what it was before.

LGTM except for a single point, addressing which may or may not lead to some changes.

yevhenii-nadtochii:
@armiol PTAL

I've updated the PR description.

model/src/main/kotlin/io/spine/validation/CompanionPolicy.kt
```diff
@@ -0,0 +1,80 @@
...
+        if (event.option.name != companionName) {
+            return noReaction
+        }
+
+        val field = event.subject
+        val primaryOption = field.findOption(primary)
+        if (primaryOption == null) {
+            Compilation.error(event.file, field.span) {
+                "The `${field.qualifiedName}` field has the `($companionName)` companion option" +
+                        " applied without its primary `($primaryName)` boolean option." +
```
alexander-yevsyukov, line 72: Why do we need to tell that the option is "boolean" in this context?

What if we have an option would be non-boolean, but still require a companion option?

alexander-yevsyukov: We don't have anything about `boolean` in the types of this class.

model/src/main/kotlin/io/spine/validation/SetOnceOption.kt
```diff
@@ -0,0 +1,127 @@
...
+}
+
+private fun checkFieldType(field: Field, file: File) =
+    Compilation.check(field.type.isSupported(), file, field.span) {
+        "The field type `${field.type.name}` of the `${field.qualifiedName}` field" +
+                " is not supported by the `($SET_ONCE)` option. Supported field types: messages," +
+                " enums, strings, bytes, bools and all numeric fields."
+    }
+
+private fun FieldType.isSupported(): Boolean = !isList && !isMap
```
alexander-yevsyukov, line 127: Please document this extension. Supported for what? Supported by whom?

model/src/test/kotlin/io/spine/validation/DistinctOptionSpec.kt
```diff
@@ -26,20 +26,22 @@
...
-    @Subscribe
-    fun on(e: DistinctFieldDiscovered) = alter {
-        errorMessage = e.errorMessage
-        subject = e.subject
+    @Test
+    fun `reject without '(distinct)'`() {
+        val message = IfHasDuplicatesWithoutDistinct.getDescriptor()
+        val error = assertCompilationFails(message)
+        val field = message.field("value")
+        error.message shouldContain field.qualifiedName
```
alexander-yevsyukov, line 43: Can we wrap the repeated `error.message` into `error.message.let { }`, please?

alexander-yevsyukov: Here and below, please.

alexander-yevsyukov (changes requested):
Nearly there. Please see my comments.

model/src/main/kotlin/io/spine/validation/SetOnceOption.kt
yevhenii-nadtochii: I've added the comment, but GitHub doesn't mark it as `outdated` for some reason.

yevhenii-nadtochii:
@alexander-yevsyukov PTAL

alexander-yevsyukov (approved):
LGTM, with the comment to address before the merge.

model/src/main/kotlin/io/spine/validation/DistinctOption.kt
```diff
@@ -88,10 +93,39 @@ internal class DistinctPolicy : Policy<FieldOptionDiscovered>() {
...
 private fun checkFieldType(field: Field, file: File) =
     Compilation.check(field.type.isSupported(), file, field.span) {
         "The field type `${field.type.name}` of `${field.qualifiedName}` is not supported" +
                 " by the `($DISTINCT)` option. This options supports `map` and `repeated` fields."
     }
 
+/**
+ * Tells if this [FieldType] can be validated with the `(distinct)` option.
+ *
+ * Returns `true` if it is supported by the option, `false` otherwise.
```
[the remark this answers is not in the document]
alexander-yevsyukov: Why didn't you use the `@returns` tag then? :)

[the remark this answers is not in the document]
yevhenii-nadtochii: This piece of docs was already there for `(required)` because it is `ineternal` there (several policies for `required` use it). I just copied it for the `private` functions.

Regarding two sentences, just documented this way from the beginning for some reason... Indeed, one of them is already enough.

PR #208, opened 2025-04-03
Declare views and policies in a single file

yevhenii-nadtochii:
This PR moves views and policies to a single file per option.

Additional changes are the following:

1. `CompilationMessageRepository` was moved to its view as well. Its name we adjusted to the name of the view state.
2. Implemented `IfInvalidPolicy` that reports a compilation warning whenever the deprecated `(if_invalid)` option is applied.
3. Moved `simpleRuleAdded()` shortcut function to `SimpleRules.kt`. It was previously located in a file named `PolicyExts`, which was not the best fit.
4. Created `BoundFieldSupport` object just to host `checkFieldType()` function. Previously, it was a top-level `internal` function. This one is similar to `RequiredFieldSupport` object.

Please note that `GoesPolicySpec` and `RangePolicySpec` (and their `*TestEnv` counterparts) retained their names. These files contain exactly one spec class for the policies.


yevhenii-nadtochii:

armiol (approved):
@yevhenii-nadtochii LGTM with a single text correction to take into account.

PR #209, opened 2025-04-03
Generate code for `(choice)` with a standalone generator

yevhenii-nadtochii:
This PR re-brands the `(is_required)` option to `(choice)`. Its code generation was migrated to a standalone generator.

Newly added classes:

1. `ChoiceGenerator` – generates code for all applications of `(choice).required = true` within a single message type.
2. `ChoiceOneofGenerator` – generates code for a particular option application.

### Naming and usage consistency

 Note that this option is for `oneof` groups, which are not fields in terms of Protobuf. 

Now, `FieldPath` (lives in `base`) is also used to represent `oneof` path. The following properties in `ConstraintViolation` are also for fields: `field_path: FieldPath` and `field_value: Any`.

As for now, we decided to postpone resolving of this inconsistency until we implement a message-wide `(require)` option. If we don't have more problems with `(require)`, I just suggest to treat `oneof` as a field within our public and internal API of validation.

### Other changes:

Removed no longer used `Diags` objects. It was meant to store error message constants, but now we store them in `options.proto` using the `(default_message)` option. Options that used this class have already been migrated to error messages defined in `options.proto`.

### ProtoData suggestions

1. `OneofGroup.qualifiedName: String` – return a fully-qualified name of `oneof`.

Note that the content of `OneofGroupExts` is going to be moved to ProtoData later.


java-tests/runtime/src/test/kotlin/io/spine/validate/option/RequiredChoiceSpec.kt
```diff
@@ -50,12 +53,14 @@ internal class IsRequiredSpec : ValidationOfConstraintTest() {
             }
         }
         val violation = exception.constraintViolations[0]
+        val oneof = Meal.getDescriptor().oneofs[0]
+        violation.fieldPath.joined shouldBe oneof.name
```
yevhenii-nadtochii, line 57: Here it is seen that I have put path to `oneof` into `fieldPath: FieldPath` of `ConstraintViolation`. 

As a simple solution, we can treat `oneof` as a field within our public and internal API. The difference is insignificant from a user point of view.

yevhenii-nadtochii:

java-runtime/src/main/kotlin/io/spine/validate/RuntimeErrorPlaceholder.kt
```diff
@@ -55,7 +55,8 @@ public enum class RuntimeErrorPlaceholder(public val value: String) {
     MAX_VALUE("max.value"),
     MAX_OPERATOR("max.operator"),
     MIN_VALUE("min.value"),
-    MIN_OPERATOR("min.operator");
+    MIN_OPERATOR("min.operator"),
+    GROUP_PATH("group.path");
```
armiol, line 59: This one is worth documenting.

java/src/main/kotlin/io/spine/validation/java/generate/option/ChoiceGenerator.kt
```diff
@@ -24,39 +24,30 @@
...
-    override fun condition(): Expression<Boolean> {
-        val casePropertyName = "${name.value}_case"
-        val pseudoField = ctx.msg.field(casePropertyName, CARDINALITY_SINGLE)
-        val getter = pseudoField.getter<Any>()
-        val numberGetter = getter.chain<Number>("getNumber")
-        return Expression("$numberGetter != 0")
+    /**
+     * All `oneof` groups with `(choice).enabled = true` in the current compilation process.
+     */
+    private val allRequiredFields by lazy {
```
armiol, line 44: In which case, this is `allRequiredGroups`.

java/src/main/kotlin/io/spine/validation/java/violation/ErrorPlaceholder.kt
```diff
@@ -62,7 +62,9 @@ public enum class ErrorPlaceholder(public val value: String) {
     MAX_VALUE("max.value"),
     MAX_OPERATOR("max.operator"),
     MIN_VALUE("min.value"),
-    MIN_OPERATOR("min.operator");
+    MIN_OPERATOR("min.operator"),
+    GROUP_PATH("group.path");
```
armiol, line 66: Similar to the one above, "group" isn't too obvious a term. Let's document.

armiol (changes requested):

yevhenii-nadtochii:
@armiol PTAL

armiol (approved):
@yevhenii-nadtochii LGTM.

PR #212, opened 2025-04-16
Generate code for the `(require)` option

yevhenii-nadtochii:
This PR implements a codegene-based validation for the `(require)` option (previously known as `(required_fields)`).

The following classes were implemented:

1. `RequirePolicy` + `RequireView`.
2. `RequireOptionGenerator` + `RequireMessageGenerator`.

This option is the last from `Workaround.kt`. The file was removed.

## `ConstraintViolation`

Note that since it is a message-wide option, the reported `ConstraintViolation` is not field-specific anymore.

## Issues

Created #213 because the provided placeholders can and should be checked during the `:model` phase, and with a compilation error pointing to a custom error message with illegal placeholders. 

In case, the `base` lib brings incompatible placeholders, `ISE` must be thrown.

#39 – will be closed with the PR. We are not going to support parentheses. Such cases were removed from tests.
#148 - will be closed with the PR as it directly addresses the issue.

PR #214, opened 2025-04-22
Move option generators and their method objects in the same file

yevhenii-nadtochii:
This PR does the following:

1. Removes `FieldOptionGenerator` interface. Now, we have `oneof` and message options as well.
2. Renames `FieldOptionCode` to `SingleOptionCode` for the same reason as stated above.
3. Moves method objects as `private` classes along with their option generators. Please note that we cannot have them all named just `Generate`. Despite the fact that they are file-private classes, there can be only one unique private class per a package in Kotlin.
4. Generators related to numeric bounds were moved under the `bound` package. We have the same package in `:model` for their policies and views.

This change set is a preparational step for #212.


yevhenii-nadtochii:

armiol:
@yevhenii-nadtochii 

> Generators related to numeric bounds were moved under the bound package. We have the same package in :model for their policies and views.

Isn't it creating a split-package situation?

alexander-yevsyukov (approved):
LGTM, provided that the split packaging issue is resolved before merging.

yevhenii-nadtochii:
> @yevhenii-nadtochii
> 
> > Generators related to numeric bounds were moved under the bound package. We have the same package in :model for their policies and views.
> 
> Isn't it creating a split-package situation?

No, everything in `:java` is prefix with `java`, and there is the `generator` subpackage too.

armiol:
@yevhenii-nadtochii So the package is same-same, but a bit different then :)

armiol (approved):

Issue #199, opened 2025-03-12
Update `EmptyFieldCheck` to avoid usage of `JavaValueConverter`


yevhenii-nadtochii:
This trait can yield expressions to check if a field holds a default value without using `JavaValueConverter` and `UnsetValue` types. 

The advantages are the following:

1. The yielded expressions would be more type-tailored. Since, we no longer rely on rules, we can check the default value not only with `equals()` method. For collections, strings and bytes, we can just check their size. For numbers, use `== 0`. And only for messages and enums, use `equals()`.
2. This trait will no longer need `UnsetValue` and `JavaValueConverter` dependencies. The latter requires `TypeSystem` to be created. The approach with creating of proto `Value`, and then converting it to Java expression looks a bit redundant when we have a `FieldType` at our disposal, and can provide `Expression` right away.

alexander-yevsyukov:
This needs to be addressed before v2.0 is out because `EmptyFieldCheck` exposes `public val converter`.

alexander-yevsyukov:
After further discussion, we can postpone this issue until after 2.0. It would require documentation update because `JavaValueConverter` is mentioned in the "Validation user guide". But it's not a big deal.

PR #215, opened 2025-04-24
Make the Java validation extendable

yevhenii-nadtochii:
This PR makes the validation library extendable for Java users. 

Now, users can create policies, views and Java generators for their custom Protobuf options.

## `ValidationOption` interface

Users can create their own providers of policies, views and Java generators by extending from the `ValidationOption` interface, and marking their implementation with `@AutoService(ValidationOption::class)`.

At the creation of `JavaValidationPlugin`, all implementations of `ValidationOption` are loaded and their views, policies and generators are being added to the plugin.

### Generator dependencies

Previously, all generators were created right within a member of a bounded context. `JavaValidationRenderer` created them passing all dependencies they needed, including those provided by the context.

Now, instances of generators must be created before they get into the context. They are firstly passed to `ValidationOption`, and later bypassed to the context member.

As for now, there are two dependencies that can be provided only by a member of a bounded context:

1. `Querying` – from our experience, every code generator needs this class to be able to fetch views, upon which the generator operators. Again, custom generators are created outside the context member. Thus, `Querying` must be injected after the generator gets into the context. For this, `OptionGenerator` is now an abstract class that has `lateinit var querying: Querying` populated later by the renderer.
2. Some generators need `EmptyFieldCheck` → `JavaValueConverter` → `TypeSystem`. The generators use `EmptyFieldCheck` trait to check arbitrary `Field`s for emptiness, which introduces a transitive dependency on `TypeSystem`, which is a member property. We can handle `TypeSystem` just as `Querying` and allow custom options to utilize `EmptyFieldCheck` as is OR address #199 because this class can be re-implemented without `TypeSystem` dependency. ProtoData events now have everything needed for codegen, the lacking information is fetched by policies. At the stage of codegen, the type system must not be used conceptually. So, I suggest addressing the mentioned issue instead of bypassing `TypeSystem` to the generators.

### Represent current options as a set of `ValidationOption` implementations

In general, it looks possible. Each option now is a combination of a generator + view + policy. 

But having them as a set of `ValidationOption`s breaks our separation of language-agnostic model and language-specific generators. This separation exists on the level of the validation library, but it is not reflected in the `ValidationOption` interface, which is Java-centric.

## `:java-api` module

All re-usable members of the `:java` and `:model` modules now reside in `:java-api` module. The `ValidationOption` interface is also part of the API module. To implement a custom option, only the API module is needed. 

## `MoneyValidationPlugin`

`MoneyValidationPlugin` that handled a custom `(currency)` option with `Rule`s became `CurrencyOption`. 

Now, it is represented with a generator, a policy and a view.


yevhenii-nadtochii:

As for now, I've made `public` only members that I needed to implement `MoneyValidationExtension`.  This is the minimum that must be `public`. I believe we should open everything from `:java` besides the generator classes (they are `final` anyway), so that the present extensions and functions could be re-used in custom generators. 

Regarding extension points. We have **one** for both language-agnostic model classes (policies, views) AND Java-specific option generators - `JavaValidationExtension`. Within the `validation` library itself, it works a bit differently. To be conceptually in-sync with the `validation` library, we must have two extension points: Java-specific (would receive custom generators) and model-specific (would receive custom context components). 

This way, there are three steps:

1. Add custom Proto option (create definition and register it).
2. Implement the option model (usually, with a policy and a view).
3. Implement generators for languages of interest (Java).

java-tests/extensions/src/main/proto/spine/validation/test/events.proto
```diff
@@ -0,0 +1,59 @@
+/*
+ * Copyright 2024, TeamDev. All rights reserved.
```
armiol, line 2: 2025.

java/src/main/kotlin/io/spine/validation/java/CustomOption.kt
```diff
@@ -0,0 +1,52 @@
...
+
+    /**
+     * The option [generator][OptionGenerator].
+     */
+    public val generator: OptionGenerator
+
+    /**
+     * The Java class of the option [view][View].
+     */
+    public val view: Class<out View<*, *, *>>
```
armiol, line 46: Let's allow having multiple `view` and `policy`.

model/src/main/kotlin/io/spine/validation/ValidationPlugin.kt
```diff
@@ -90,5 +96,5 @@ public abstract class ValidationPlugin(renderers: List<Renderer<*>> = emptyList(
         SetOncePolicy(),
         IfSetAgainPolicy(),
         RequirePolicy()
-    )
+    ) +  policies
```
armiol, line 99: Please review this file for code formatting.

model/src/main/kotlin/io/spine/validation/ValidationPlugin.kt
```diff
@@ -51,7 +52,12 @@ import io.spine.validation.required.RequiredPolicy
  * The concrete implementations should provide [renderers], which implement
  * these constraints for a specific programming language.
  */
-public abstract class ValidationPlugin(renderers: List<Renderer<*>> = emptyList()) : Plugin(
+public abstract class ValidationPlugin(
```
armiol, line 55: Having this type as `public` is now confusing. Given we don't suggest that end-users create their own plugins, but rather extend via `CustomOption`, this type should be `@Internal` at the very least.

java/src/main/kotlin/io/spine/validation/java/generate/OptionGenerator.kt
```diff
@@ -27,17 +27,34 @@
...
      *
      * @param type The message to generate code for.
      */
-    fun codeFor(type: TypeName): List<SingleOptionCode>
+    public abstract fun codeFor(type: TypeName): List<SingleOptionCode>
+
+    /**
+     * Injects [Querying] into this instance of [OptionGenerator].
+     */
+    internal fun inject(querying: Querying) {
```
armiol, line 57: Let's mark this `@Internal`.

armiol (changes requested):
@yevhenii-nadtochii please see the first-round comments.

I also think we need a separate module (and artifact) which clearly declares the public API for SPI users. Once this is done, we will need to update the documentation with this regard, and mark some types as `@SPI`.

As discussed, we will meet today on the call, and talk over the remaining points.

armiol:
@yevhenii-nadtochii I am a bit confused with `CustomOption` mentioned in the description, while we have recently discussed it to become `ValidationOption`.

armiol:
@yevhenii-nadtochii I also don't really understand the point of merging this PR as-is. Wouldn't it be easier to remove the rules in `master`, and then back-merge the changes to this PR? 

Yes, you will have this test disabled in that "remove rules" PR, but they don't make any sense right now anyway.

armiol:
@yevhenii-nadtochii is this a draft or something ready for reviewing?

yevhenii-nadtochii:
> @yevhenii-nadtochii is this a draft or something ready for reviewing?

This is a draft.

model/src/main/kotlin/io/spine/validation/ValidationPlugin.kt
yevhenii-nadtochii: It was `public` before, I haven't changed its visibility.

It is `public` for consumers of the `:model` module, which are language-specific plugins.

yevhenii-nadtochii:

armiol (approved):
@yevhenii-nadtochii LGTM.

In the following rounds we'll need to review the newly created module for clarity. However, it is easier to do in the IDE rather than via the GitHub PR review.

In the meantime, could you please append the repo-level `README.md` with some notes on how to provide own custom options.

Issue #213, opened 2025-04-18
Make policies check the error message placeholders


yevhenii-nadtochii:
Currently, in case of an invalid placeholder, `IllegalArgumentException` is thrown in runtime.

We should teach our policies to validate the used placeholders and report illegal ones with a compilation error.

PR #216, opened 2025-04-25
Report a compilation error for invalid placeholders [Part I]

yevhenii-nadtochii:
This PR addresses #213 for boolean options that use a companion option to specify custom error messages.

The following options have received this compilation check: `(distinct)`, `(required)`, `(set_once)`.

Their companion policies now extract and check the error messages on their own. Instead of a single event, the main and companion policies now emit their own events that are accumulated in the view:

1. `RequiredFieldDiscovered` → `RequiredFieldDiscovered` and `IfMissingOptionDiscovered`.
2. `DistinctFieldDiscovered` → `DistinctFieldDiscovered` and `IfHasDuplicatesOptionDiscovered`.
3. `SetOnceFieldDiscovered` → `SetOnceFieldDiscovered` and `IfSetAgainOptionDiscovered`.

Note that we usually emit the following event: `${OptionName}{OptionTarget}Discovered`. I kept this convention for the main options, but emitting something like `IfMissingFieldDiscovered` makes no sense because this is actually an option that affects another, its main option, not the field itself.

Having these checks in `:model` required the following changes:

1. The `ErrorPlaceholder` enum has been moved to `:model`.
2. `resolveErrorMessage()` function is no longer needed. Companion policies now unpack the option and extract the message in a typed manner.
3. `CompanionPolicy` has been replaced in favor of the `checkBothApplied()` function.

As for now, I haven't extracted a common `checkPlaceholders()` function to check for placeholders. I'll do in the next PR when the rest of options will be given such a check.

## Non-related changes

`ChoiceView` has been renamed to `ChoiceGroupView` to fix naming consistency: `${OptionName}{OptionTarget}View`.


yevhenii-nadtochii:

model/src/main/kotlin/io/spine/validation/CompanionOptions.kt
```diff
@@ -0,0 +1,58 @@
...
+/**
+ * Reports a compilation error if the [companion] option is applied
+ * to the given [field] without the [primary] option.
+ *
+ * Some options have a companion option for specifying an error message.
+ * This method ensures that a companion option is not used independently.
+ */
+internal fun checkBothApplied(
+    companion: GeneratedExtension<*, *>,
+    primary: GeneratedExtension<*, *>,
```
armiol, line 46: Instead of having a function taking two _very_ similar arguments (in which I would have the `primary` first, BTW), I'd rather have an extension function to `GeneratedExtension`.

The reason is that  the invocations of `callFunction(someparameter, someotherparameter)` in which both of the arguments are of the same type, are prone to errors with the parameters swapped.

armiol (changes requested):
@yevhenii-nadtochii LGMT in general. Please see my comment though.

yevhenii-nadtochii:
@armiol PTAL

armiol (approved):

PR #223, opened 2025-05-05
Update README

yevhenii-nadtochii:
This PR actualizes README file, mainly by adding three sections:

1. Overview.
2. Structure.
3. Customizing.

Note that there is no `Getting Started` section as for now. It should be added when the library becomes a Gradle plugin.

I have also removed `compileOnly(AutoService.annotations)` dependency from `:extensions` module because it is not needed when `ksp(AutoServiceKsp.processor)` is present.


yevhenii-nadtochii:
@armiol PTAL

README.md
```diff
@@ -2,14 +2,183 @@
...
 
-# Spine Validation SDK
+# Spine Validation
 
-This repository contains Spine Validations library and development tools for working with 
-data types validation.
+The library brings data validation directly into your Protobuf messages.
 
-## Java Support
+### Why Use
```
armiol, line 9: Just "Why".

README.md
```diff
@@ -2,14 +2,183 @@
...
 
-This library is built with Java 11.
+Protobuf messages are defined using a language that lacks built-in mechanisms for enforcing 
+domain-specific rules at runtime. As a result, developers must manually insert validation logic
```
armiol, line 12: "must" is too strong a word.

I would first say why we need validation as an idea. And then say how they would do it without this library.

armiol (approved):
@yevhenii-nadtochii See my comments.

Other than that, LGTM for the first round. We'll be modifying this text in the future anyway.

PR #224, opened 2025-05-06
Support custom message validators

yevhenii-nadtochii:
This PR brings support for the custom `MessageValidator`s.

The feature specification resides in the docs to the `MessageValidator` interface.

This feature is based around two main entities residing in the `:java-api` module:

1. `MessageValidator<T : Message>` – an interface that defines behavior of the validators.
2. `@Validator(T::class)` – an annotation used for the discovering of the validators by the library.

The following modules were added:

1. `:java-ksp` – responsible for discovering of the annotated validators. The module outputs the discovered validators to a plain text file within the task output directory of KSP.
2. `:java-tests:validator-dependency` – declares an external message for tests.
3. `:java-tests:validator` – tests validators with external and well-known Protobuf messages.

`JavaValidationPlugin` fetches the file left by the KSP task, and passes the class names of the validators and their target messages to the renderer. The renderer bypasses them to  `ValidatorGenerator`, which generates the actual invocations.

Note that `ValidatorGenerator` does not belong to the hierarchy of the option generators. It stands on its own for the following reasons:

1. Conceptually, it has nothing to do with the options.
2. It has to accept `MessageType` instead of `TypeName` because it needs to go through the message fields.

## Additional changes

Added an extension `fun <T : Dependency> T.artifact(module: T.() -> String): String`, which allows using the properties of the target `Depenency` object without referencing the object itself.


yevhenii-nadtochii:

java-api/src/main/kotlin/io/spine/validation/api/MessageValidator.kt
```diff
@@ -0,0 +1,129 @@
...
+
+package io.spine.validation.api
+
+import com.google.protobuf.Message
+import io.spine.annotation.SPI
+
+/**
+ * A custom validator for Protobuf messages of type [M] that are defined externally.
+ *
+ * External messages are those that come from dependencies.
```
armiol, line 35: We consider "external" those `Message` definitions for which our end-users do not generate the Java/Kotlin code.

The definition you give is just misleading because end-users can generate code for any Proto type, including those coming from any third-party dependency. Many libraries come with their own generated and compiled Java code for their Proto definitions (such as Google Cloud libraries). And it is pretty difficult to generate another version of Java code and still make the libraries work properly.

I think the beginning (maybe even including the first paragraph) should be updated accordingly. Let's consider the readers much less prepared.

java-api/src/main/kotlin/io/spine/validation/api/MessageValidator.kt
```diff
@@ -0,0 +1,129 @@
...
+ * public class EarphonesValidator : MessageValidator<Earphones> {
+ *     public override fun validate(message: Earphones): List<DetectedViolation> {
+ *         return emptyList() // Always valid.
+ *     }
+ * }
+ * ```
+ *
+ * ## Applicability
+ *
+ * A validator is applied only to the local messages of the module it is declared in.
```
armiol, line 58: I don't understand what you are trying to say here.

java-api/src/main/kotlin/io/spine/validation/api/MessageValidator.kt
```diff
@@ -0,0 +1,129 @@
...
+ *
+ * Please note that standalone instances of [M] and fields of [M] type that occur in
+ * external messages will not be validated using the validator.
+ *
+ * Consider the following example:
+ *
+ * ```proto
+ * import "earphones.proto"; // Brings the `Earphones` message from dependencies.
+ *
+ * // A locally-declared message.
```
armiol, line 76: See, the problem with this "local-external" is that they are not defined properly. Please devote a section in the beginning to the explanations and the definitions.

java-api/src/main/kotlin/io/spine/validation/api/MessageValidator.kt
```diff
@@ -0,0 +1,132 @@
...
+import io.spine.annotation.SPI
+
+/**
+ * A custom validator for an external Protobuf message of type [M].
+ *
+ * ## Notation of external and local messages
+ *
+ * **Local messages** are message types for which end-users generate and control
+ * the Java/Kotlin classes by compiling the corresponding `.proto` definitions
+ * within their own codebase. For such messages, the validation library allows
```
armiol, line 39: "Validation". It's a name.

java-api/src/main/kotlin/io/spine/validation/api/MessageValidator.kt
```diff
@@ -0,0 +1,132 @@
...
+
+package io.spine.validation.api
+
+import com.google.protobuf.Message
+import io.spine.annotation.SPI
+
+/**
+ * A custom validator for an external Protobuf message of type [M].
+ *
+ * ## Notation of external and local messages
```
armiol, line 35:
```
"## "External" and "local" `Message`s types
```

java-api/src/main/kotlin/io/spine/validation/api/MessageValidator.kt
```diff
@@ -0,0 +1,132 @@
...
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+package io.spine.validation.api
+
+import com.google.protobuf.Message
+import io.spine.annotation.SPI
+
+/**
+ * A custom validator for an external Protobuf message of type [M].
```
armiol, line 33: > A custom validator ...

Do we have our own implementations of this `MessageValidator`? I mean, is it expected that all implementations are for the custom "external" `Message` types only?

java-api/src/main/kotlin/io/spine/validation/api/MessageValidator.kt
```diff
@@ -0,0 +1,132 @@
...
+ *
+ * ## Notation of external and local messages
+ *
```
armiol, line 36: Before explaining which type is which, we need to have an introduction to create some context for the reader. We need to explain what is the problem.

Usually, Java libraries which are based off some Proto types, bring these types with them, not only as `.proto` definitions, but also as compiled `.class` files. 

Let's have a section even before this one explaining the problem itself. Maybe, with this new section, the one for "Notation" will be transformed or even go away completely.

armiol (changes requested):
@yevhenii-nadtochii please see more comments on the same matter.

yevhenii-nadtochii:

java-api/src/main/kotlin/io/spine/validation/api/MessageValidator.kt
```diff
@@ -0,0 +1,141 @@
...
+ * Please keep in mind that for each invocation a new instance of [MessageValidator]
+ * is created. Every implementation of [MessageValidator] must have a public,
+ * no-args constructor.
+ *
+ * An implementation of [MessageValidator] will be rejected by the library
+ * in the following cases:
+ *
+ * 1) It is used to validate a local message. Only external messages are allowed
+ *    to have a validator.
+ * 2) There already exists a validator for the specified message type. Having several
```
armiol, line 127: Let's also describe the scope in which these rules apply. E.g. one may think it's scoped by a Gradle module, while it is certainly not.

Please have it somewhere above, as it is an important point.

java-api/src/main/kotlin/io/spine/validation/api/MessageValidator.kt
```diff
@@ -0,0 +1,141 @@
...
+ * }
+ * ```
+ *
+ * Supposing that the Validation library applied to the module where both `WorkingSetup`
+ * and `EarphonesValidator` classes are declared, then the generated code of `WorkingSetup`
+ * will apple the validator to each instance passed to the `WorkingSetup.earphones` field.
+ *
+ * ## Implementation
+ *
+ * The message validator does not have restrictions upon how exactly the message
```
armiol, line 108: Let's start with a positive sentence, not with the one built around the negation.

armiol (changes requested):
@yevhenii-nadtochii getting closer. Please see my comments.

alexander-yevsyukov (approved):
LGTM

yevhenii-nadtochii:
@armiol PTAL

PR #225, opened 2025-05-07
Use field references for numeric bounds

yevhenii-nadtochii:
This PR adds support for field references in `(min)`, `(max)` and `(range)` options. 

The feature specification was described [here](https://github.com/SpineEventEngine/base/pull/877).

`NumericBound` was extended to support `FieldPath` as another value case for its `oneof` field. Policies and generators were updated to parse and render code for the field paths.

## ProtoData suggestions

I suggest moving the content of `StringExpressions.kt` to ProtoData.

Namely:

1. `Expression<String>.plus(value: String): Expression<String>` – concatenates `String` expression and `String`.
2. `Expression<String>.plus(value: Expression<String>): Expression<String>` – concatenates two `String` expressions.

## Unsigned integer warnings for Java

 I've temporarily [disabled](https://github.com/SpineEventEngine/validation/issues/227) them until we are sure how to proceed with them.

Issue #220, opened 2025-05-02
Align warning message in console for better readability


alexander-yevsyukov:
When building Validation with latest dependencies I got the following console output. I'm placing the screenshot with slightly shrinked IDEA app window to hightlight the matter:

<img width="1416" alt="Image" src="https://github.com/user-attachments/assets/9b0cac1c-c5e8-4a10-b12c-2bd04ed3747c" />
 
The text we are puttin into the console is this:

```
> Task :java-tests:runtime:launchTestProtoData
w: file:///Users/sanders/Projects/Spine/validation/java-tests/runtime/src/test/proto/spine/test/validation/range_option_test.proto:41:5: Unsigned integer types are not supported in Java. The Protobuf compiler uses signed integers to represent unsigned types in Java (https://protobuf.dev/programming-guides/proto3/#scalar). Operations on unsigned values rely on static utility methods from `java.lang.Integer` and `java.lang.Long` classes (https://www.baeldung.com/java-unsigned-arithmetic#the-unsigned-integer-api). Be cautious when dealing with unsigned values outside of these methods, as Java treats all primitive integers as signed.
w: file:///Users/sanders/Projects/Spine/validation/java-tests/runtime/src/test/proto/spine/test/validation/range_option_test.proto:54:5: Unsigned integer types are not supported in Java. The Protobuf compiler uses signed integers to represent unsigned types in Java (https://protobuf.dev/programming-guides/proto3/#scalar). Operations on unsigned values rely on static utility methods from `java.lang.Integer` and `java.lang.Long` classes (https://www.baeldung.com/java-unsigned-arithmetic#the-unsigned-integer-api). Be cautious when dealing with unsigned values outside of these methods, as Java treats all primitive integers as signed.
```

It would be better to wrap the text of one warning with proper indentation. It would make these warnings:
 1. Easier to read and understand. The user won't have to force his eyeballs travel the whole width of the screen back and forth.
 2. Easier to navigate. Indentation would form `w: ` prefixes as bulleted items.

It should look like this:

```
> Task :java-tests:runtime:launchTestProtoData
w: file:///Users/sanders/Projects/Spine/validation/java-tests/runtime/src/test/proto/spine/test/validation/range_option_test.proto:41:5: Unsigned integer types are not supported in Java.
   The Protobuf compiler uses signed integers to represent unsigned types in Java.
   See: https://protobuf.dev/programming-guides/proto3/#scalar
[...]

PR #228, opened 2025-05-13
Refactor parsing of numeric options

yevhenii-nadtochii:
This PR is a follow-up for recently introduced [field references](https://github.com/SpineEventEngine/validation/pull/225). It aims to refactor parsing of numeric options to bring more clarity because the code became a little more complicated.

1. In test stubs, I have renamed one of the fields to begin with the underscore to make sure it works as well.
2. `KotlinNumericBound` was renamed to `KNumericBound`.
3. Extracted a standalone `NumericBoundParser`. Now we have a lot of logic here (3 methods), so it cannot be just an auxiliary function in `KNumericBound.kt`.
4. `BoundContext` has been renamed to `NumericOptionMetadata`. The parser accepts it on the constructor.
5. Minor refactorings of string-parsing code.
6. Reformatted error messages to be multi-line. Partially addresses #220. 

## `spine-base` suggestions

I suggest move to the `base`:

```kotlin
/**
 * Returns the simple class name of the given [any].
 */
public fun simpleNameOf(any: Any): String = any::class.simpleName!!
```


yevhenii-nadtochii:

model/src/main/kotlin/io/spine/validation/bound/BoundDetails.kt
```diff
@@ -28,37 +28,49 @@ package io.spine.validation.bound
...
 import io.spine.protodata.ast.File
-import io.spine.protodata.ast.MessageType
 import io.spine.protodata.ast.PrimitiveType
 import io.spine.protodata.type.TypeSystem
 import io.spine.validation.RANGE
 
 /**
- * The context of validating a numeric option that constrains a field's value
- * with a minimum or maximum bound.
+ * A container for auxiliary data required to parse and validate
```
armiol, line 36: The problem with this documentation is that it says "it's a bag of things related to a numeric bound".

Please try to give a more specific answer to "what it is" in this documentation. Then, it will be easier to give the name.

Just for the reference, my idea (prior to your PR) was `BoundDeclaration`, roughly meaning "the `bound` option declared <somewhere>"?

armiol (changes requested):

model/src/main/kotlin/io/spine/validation/bound/KotlinNumericBound.kt
```diff
@@ -140,60 +140,60 @@ internal fun KotlinNumericBound.toProto(): NumericBound {
...
 ): KotlinNumericBound {
-    Compilation.check(value.isNotEmpty(), file, field.span) {
+    Compilation.check(bound.isNotEmpty(), file, field.span) {
         "The `($optionName)` option could not parse the bound value specified for" +
                 " `${field.qualifiedName}` field because it is empty. Please provide either" +
                 " a numeric value or a field reference."
     }
-    return if (value.first().isLetter()) {
-        checkFieldValue(value, exclusive)
+    return if (bound.first().isLetter()) {
```
alexander-yevsyukov, line 152: This reads a bit cryptic. Do we mean here that `bound.isReference()`?

yevhenii-nadtochii:
@armiol PTAL

I've moved from `how it is used` to `what it is`: `BoundDetails` → `NumericOptionMetadata`.

>> `BoundDeclaration`, roughly meaning "the bound option declared [somewhere]".

I'm not really sure about `declaration`. Declaration of what? This bag contains `file, typeSystem, fieldType`.

It can also be sort of `Descriptor` for the option declaration. But this name is used by the Protobuf API. Naming like `BoundOption`, `RangeOption` clashes with the name of Protobuf-generated option definitions: `MaxOption, MinOption, RangeOption`.

model/src/main/kotlin/io/spine/validation/bound/RangeOption.kt
```diff
@@ -109,18 +109,17 @@ internal class RangePolicy : Policy<FieldOptionDiscovered>() {
...
+        val optionMetadata = RangeOptionMetadata(option.value, field, fieldType, file, typeSystem)
+        val delimiter = optionMetadata.checkDelimiter()
 
-        val (left, right) = context.range.split(delimiter)
-        val (lowerExclusive, upperExclusive) = context.checkBoundTypes(left, right)
-        val lower = context.checkNumericBound(left.substring(1), lowerExclusive)
-        val upper = context.checkNumericBound(right.dropLast(1), upperExclusive)
+        val (left, right) = optionMetadata.range.split(delimiter)
+        val (lowerExclusive, upperExclusive) = optionMetadata.checkBrackets(left, right)
+        val lower = optionMetadata.checkNumericBound(left.substring(1), lowerExclusive)
```
alexander-yevsyukov, line 117: What are those `left.substring(1)` and `right.dropLast(1)` are about?

model/src/main/kotlin/io/spine/validation/bound/RangeOption.kt
```diff
@@ -109,18 +109,17 @@ internal class RangePolicy : Policy<FieldOptionDiscovered>() {
         val fieldType = checkFieldType(field, file, RANGE)
 
         val option = event.option.unpack<RangeOption>()
-        val (messageType, _) = typeSystem.findMessage(field.declaringType)!!
-        val context = RangeContext(option.value, typeSystem, messageType, fieldType, field, file)
-        val delimiter = context.checkDelimiter()
+        val optionMetadata = RangeOptionMetadata(option.value, field, fieldType, file, typeSystem)
+        val delimiter = optionMetadata.checkDelimiter()
```
alexander-yevsyukov, line 113: Why do we need two steps for splitting the range? We don't seem to use the `delimiter` outside the calls to `optionMetadata`. 

Can we encapsulate it somehow, please?

model/src/main/kotlin/io/spine/validation/bound/KotlinNumericBound.kt
```diff
@@ -121,7 +121,7 @@ internal fun KotlinNumericBound.toProto(): NumericBound {
 }
 
 /**
- * Parses the given string [value] to a [KotlinNumericBound].
+ * Parses the given raw [bound] string to a [KotlinNumericBound].
```
alexander-yevsyukov, line 124: I cannot comment on the above code, so I do it here. What's the value of having the `Kotlin` prefix of this class name? We are already under the `kotlin` directory. Everything around is written in Kotlin. The class is `internal`.

Why do we need to stress on the fact that this class uses unsigned types from Kotlin? Are we shy about it, or what?

alexander-yevsyukov (changes requested):

model/src/main/kotlin/io/spine/validation/bound/KotlinNumericBound.kt
armiol, line 124: @yevhenii-nadtochii with this documentation, this is `NumericOptionParser`.

model/src/main/kotlin/io/spine/validation/bound/RangeOption.kt
yevhenii-nadtochii: So that to avoid passing bound values with inclusive/exclusive brackets `][)(`.

Passing with brackets would have forced the parser to check for `([` at the beginning and `)]` at the end. 

Min/max options also use this method, they don't have such symbols at all.

So, the `(range)` option has its own `optionMetadata.checkBrackets(left, right)` compilation check.

model/src/main/kotlin/io/spine/validation/bound/KotlinNumericBound.kt
yevhenii-nadtochii: >> What's the value of having the Kotlin prefix of this class name?

Because it is a one-to-one copy of Protobuf-generated `NumericBound`, but with Kotlin types used to represent a number (instead of primitives). This simplifies parsing and comparison (the latter is needed further to check the relation of the both range bounds).

yevhenii-nadtochii:

model/src/main/kotlin/io/spine/validation/bound/NumericBoundParser.kt
```diff
@@ -0,0 +1,203 @@
...
+import io.spine.protodata.type.resolve
+import io.spine.validation.bound.BoundFieldSupport.numericPrimitives
+
+/**
+ * Parses bound values specified in Protobuf options that restrict
+ * the value range of numeric fields.
+ *
+ * The parser supports both number literals and field references.
+ *
+ * @param metadata the metadata of the numeric option, for which the bound is parsed.
```
alexander-yevsyukov, line 55: We agreed to document parameters as English sentences. So, please capitalise `The` before `metadata`. 

Also, if possible, please think of rephrasing the description avoiding "Metadata is metadata" thing.

model/src/main/kotlin/io/spine/validation/bound/NumericBoundParser.kt
```diff
@@ -0,0 +1,203 @@
...
+/**
+ * Tells whether this [String] bound value contains a field reference.
+ */
+private fun String.isFieldReference() = first().run {
+    isLetter() || this == '_'
+}
+
+private fun unexpectedPrimitiveType(primitiveType: PrimitiveType): Nothing =
+    error(
+        "`KotlinNumericBound` cannot be created for `$primitiveType` field type." +
```
alexander-yevsyukov, line 201: Is it still `KotlinNumericBound`? Can we have it as just `NumericBound`, please?

alexander-yevsyukov (changes requested):
Please see my comments.

model/src/main/kotlin/io/spine/validation/bound/RangeOption.kt
alexander-yevsyukov: It is better to have these explanations "embedded" in the code, rather than left as a comment to a PR. I'll try to explain it differently.

The code like this `left.substring(1)` looks mysterious. We cut the first character in something `left`. Why? For what? It is better to have a variable which explains the result of the operation. It would help understanding the code of parsing, which is arguably one of the complicated things we do in this project.

model/src/main/kotlin/io/spine/validation/bound/KotlinNumericBound.kt
alexander-yevsyukov: Can we have it as `KNumericBound` then? As a last resort.

yevhenii-nadtochii:

armiol (approved):

model/src/main/kotlin/io/spine/validation/bound/NumericBoundParser.kt
```diff
@@ -0,0 +1,201 @@
...
+     *
+     * Any violation of the above conditions leads to a compilation error.
+     *
+     * @return The parsed numeric bound.
+     */
+    internal fun parse(bound: String, exclusive: Boolean): KNumericBound {
+        with(metadata) {
+            Compilation.check(bound.isNotEmpty(), file, field.span) {
+                "The `($optionName)` option could not parse the bound value specified for" +
+                        " `${field.qualifiedName}` field because it is empty. Please provide" +
```
alexander-yevsyukov, line 86: Please have the ` Please provide` staring on the new line in the code, and in the console output.
This way it's easier to read.

alexander-yevsyukov (changes requested):
Please see my comments.

model/src/main/kotlin/io/spine/validation/bound/RangeOption.kt
```diff
@@ -109,18 +109,21 @@ internal class RangePolicy : Policy<FieldOptionDiscovered>() {
...
+        val metadata = RangeOptionMetadata(option.value, field, fieldType, file, typeSystem)
+        val delimiter = metadata.checkDelimiter()
 
-        val (left, right) = context.range.split(delimiter)
-        val (lowerExclusive, upperExclusive) = context.checkBoundTypes(left, right)
-        val lower = context.checkNumericBound(left.substring(1), lowerExclusive)
-        val upper = context.checkNumericBound(right.dropLast(1), upperExclusive)
+        val (left, right) = metadata.range.split(delimiter)
+        val (lowerExclusive, upperExclusive) = metadata.checkBrackets(left, right)
+        val leftWithoutBracket = left.substring(1)
```
alexander-yevsyukov, line 117: `leftWithoutBracket` is probably `lowerBound` or something like that.

java-api/src/main/kotlin/io/spine/validation/api/Strings.kt
```diff
@@ -0,0 +1,32 @@
...
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+package io.spine.validation.api
+
+/**
+ * Returns the simple class name of the given [any].
```
alexander-yevsyukov, line 30: `given [instance].`?

alexander-yevsyukov (approved):
LGTM, with one minor request.

PR #232, opened 2025-05-20
Use compact JSON for `Message` placeholders

yevhenii-nadtochii:
This PR updates codegen so that Protobuf messages are printed with `toCompactJson()` whenever they are used as placeholder values. 

For the other types, `toString()` and `String.valueOf()` are still used.

Please note, some options cannot longer use `FieldType.stringValueOf()` extension because they report an error for `repeated` fields using a single value. Hopefully, values in such options are restricted by type, so it is always clear how to stringify them. The options are the following:

1. `(min)`, `(max)` and `(range)` – they always operate upon integers, so `String.valueOf()` is always used. When such an option applied to `repeated`, each list value is asserted and reported **individually**. So, we cannot use `FieldType.stringValueOf()` because it will try to stringify a singular value as a list. There is a mismatch between a field type and reported value.
2. `(when)` – the same case. When being applied to `repeated`, it asserts and reports individual values intead of the whole collection.

It happened to work previously because invocations of `list.toString()` and `value.toString()` made no difference when `value` is a boxed primitive. All numeric values are boxed when they are used within a collection. So, we had no problems here. But now we invoke `stream()` for collections in `FieldType.stringValueOf()`, and this "by chance" workable approach has failed.

Additional changes:

1. `DistinctGenerator` now puts a map of duplicates for `map` fields, so that the keys of duplicates are also visible. This can be important because keys are unique. As a bonus, this change alligns the type of the `duplicates` placeholder value with the field type of the option, making possible to use `FieldType.stringValueOf()` for it.
2. Migrated usages of some local extensions to the recently moved counterparts in `base` and `ProtoData`.
4. Made error messages of `KNumericBound` more self-descriptive.


yevhenii-nadtochii:

model/src/main/kotlin/io/spine/validation/bound/KNumericBound.kt
```diff
@@ -95,8 +104,12 @@ internal fun KNumericBound.toProto(): NumericBound {
         is FieldPath -> builder.setFieldValue(value)
 
         else -> error(
-            "Cannot convert `${simply<KNumericBound>()}` to `${simply<NumericBound>()}`" +
-                    " due to unexpected value type: `${simpleNameOf(value)}`."
+            """
+            Cannot convert `${qualified<KNumericBound>()}` to its Protobuf counterpart: `${simply<NumericBound>()}`.
+            `${simply<KNumericBound>()}` contain a value with unsupported type.
```
armiol, line 109: "contains"?

armiol (changes requested):

yevhenii-nadtochii:

alexander-yevsyukov (approved):
LGTM

PR #293, opened 2026-05-05
Developer's guide

alexander-yevsyukov:
This PR adds the “Validation developer guide” to the documentation. The navigation of the preview site was changed so that "User guide" and "Developer guide" appear as top-level nodes on the left side panel. This, in turn, required having `user` and `developer` subdirectories under `docs/validation/`.

This PR also includes latest updates from `site-commons` that bring improved table styes and page level navigation on the right.

### Other notable changes
 * The `/writer` skill was updated to use typographic quotes when referencing sections and pages in the text body.
 * `embed-code.yml` was updated with the `root` code path, `code-includes` parameter was also added to avoid copying files under `.git/` which failed the build. It will also make the embedding work a bit faster because we won't copy anything but embedded content candidates.


docs/_preview/layouts/_partials/theme/functions/get-doc-section.html
```diff
@@ -0,0 +1,46 @@
+<!--
+  Returns the section of the documentation file used to match sidenav entries.
+
+  Finds the longest matching `content_path` from `data/versions.yml` and returns
```
alexander-yevsyukov: @JuliaEvseeva, please have a look at the update to the logic of selecting a section. I had to apply it (via Codex) so that User's guide and Developer's guide sections can work as top level siblings in navigation.

Now the fix is applied to the preview site only. It should probably to go `site-commons` so that we can use it in Validation and other libraries.

This is a minimal change for now, from what I see. We should probably work more on the styling for navigation so that it is possible to have "Validation Library" on the top level as a grand section and then "User guide" and "Developer guide" as nested subsections with following nesting.

alexander-yevsyukov: @armiol, FYI.

JuliaEvseeva: The logic will be copied to the `site-commons` in the [#28](https://github.com/SpineEventEngine/site-commons/pull/28). I have checked the versions, and the other sidenav sections will work as expected.

alexander-yevsyukov:
@armiol, it accidentally merged because of misconfiguration of the branch rule settings. I'll undo the PR, and create a new one.

PR #315, opened 2026-06-01
Update config, migrate to Kover

alexander-yevsyukov:
This PR applies latest `config` and migrates to Kover. Several minor tests were also added under `jvm-runtime` module.


build.gradle.kts
```diff
@@ -73,8 +73,8 @@ buildscript {
 }
 
 plugins {
+    base
```
alexander-yevsyukov, line 76: @copilot, do we really need to apply the `base` plugin here?

alexander-yevsyukov (commented):
@copilot, please see my comments and questions.

jvm-runtime/src/test/kotlin/io/spine/validation/MessageExtensionsSpec.kt
```diff
@@ -0,0 +1,148 @@
...
+import com.google.protobuf.StringValue
+import com.google.protobuf.UnknownFieldSet
+import io.kotest.matchers.shouldBe
+import io.spine.base.FieldPath
+import io.spine.type.TypeName
+import java.util.*
+import org.junit.jupiter.api.DisplayName
+import org.junit.jupiter.api.Test
+
+@DisplayName("message validation extensions should")
```
armiol: Let's at least capitalize the display name.

jvm-runtime/src/test/kotlin/io/spine/validation/MessageExtensionsSpec.kt
```diff
@@ -0,0 +1,148 @@
...
+internal class MessageExtensionsSpec {
+
+    @Test
+    fun `delegate 'checkValid' to 'Validate_check'`() {
```
armiol, line 47: How about we use `->` in such cases? E.g. `delegate 'checkValid' to 'Validate->check'`?

I remember, there is even a better arrow, but it's more difficult to type: `→` :)
`delegate 'checkValid' to 'Validate→check'`

armiol (changes requested):
@alexander-yevsyukov LGTM in general. Please see my comments, though.

jvm-runtime/src/test/kotlin/io/spine/validation/MessageExtensionsSpec.kt
alexander-yevsyukov: We cannot have `>` in a function name. The `→` causes warning.

armiol (approved):

PR #317, opened 2026-07-06
Credit Spine Compiler codegen execution to coverage

alexander-yevsyukov:
## What & why

The `java` and `context` modules — the Spine Compiler plugins, renderers, and option
generators — execute in the **forked `launch*SpineCompiler` JVM** during the build, which
Kover/JaCoCo do not instrument. They reported **~0% line coverage** even though the `tests/*`
`.proto` fixtures exercise them on every build, holding the repo total near **12%** (matching
Codecov).

This attaches the standalone JaCoCo agent to the forked compiler JVM, harvests its `.exec`, and
merges it into the **root** Kover report (the one Codecov ingests) — mirroring the existing
`enableTestKitCoverage()` pipeline for Gradle TestKit workers.

## How

- **`enableSpineCompilerCoverage()`** (new `buildSrc/.../testing/SpineCompilerCoverage.kt`)
  attaches `-javaagent:<Jacoco.agent>=destfile=…/<task>.exec,append=false` to every
  `launch*SpineCompiler` `JavaExec` task, with a per-task destfile.
- The `.exec` is **declared as a task output**, so the launch tasks stay **cacheable**: a
  `JavaExec` blocks until the fork exits and the agent flushes the exec on exit, so the file is on
  disk when the task action returns and *can* be a declared output — unlike a lingering TestKit
  worker daemon (which is why `enableTestKitCoverage()` still disables caching). Coverage therefore
  survives both an `UP-TO-DATE` skip and a `FROM-CACHE` hit.
- **`KoverConfig`** feeds the exec into the root report's `additionalBinaryReports`, and orders
  `kover*Report`/`kover*Verify` after the launch tasks — the root report does not otherwise depend
  on the non-Kover `tests/*` modules that produce the exec.
- The JaCoCo agent coordinate is centralized as `Jacoco.agent` (shared with the TestKit helper).
- Applied via `subprojects { enableSpineCompilerCoverage() }`; a no-op where no launch task exists.

Class IDs match by construction: the compiler CLI shadow jar does no `relocate()` and loads
consumer plugin classes as their original artifacts (verified in the `compiler` repo, Phases 0–1).

## Result (verified locally)

`./gradlew clean build koverXmlReport` → root report **12% → 86.3%** line coverage
(`covered 1894 / 2195`), reconciling exactly with Codecov's 12.35% baseline. Ten `.exec` files are
produced across all three launch-task variants. Generators flip 0% → covered:
`MinGenerator`/`MaxGenerator`/`RangeGenerator` 100%, `JavaValidationRenderer` 97.8%,
`BoundedFieldGenerator` 88%. `ValidatorGenerator` stays **0%** — a genuine gap (no fixture declares
a custom `@Validator`), confirming the measure is discriminating rather than blanket-crediting.

**Caching verified:** a no-change re-run is `UP-TO-DATE`; deleting the exec makes the task
`FROM-CACHE` and restores the exec; the full build reports `1 from cache`.

## Caveats / follow-ups

- This is **execution coverage** of codegen, not assertion-backed unit coverage — genuine (the code
  runs against the fixtures) but more generous. A follow-up should report it under a **separate
  Codecov flag** (`codegen` vs `runtime`) and upstream the two `buildSrc` helpers to `config`.
- The version bump's dependency/license-report refresh was not run separately; a report-update
  commit from the working tree is included, but if CI's report check flags a `.448` mismatch, run
  the report task.

## Pre-PR

PASS — `./gradlew build dokkaGenerate` green; `spine-code-review` + `kotlin-engineer` APPROVE.

## Note on commits

Opened as-is per request. Beyond the coverage feature, the version bump, and a review-feedback
commit (cacheable exec + `Jacoco.agent`), this PR also carries three commits from local-`master`
drift: a team-memory note (`.agents/memory/*`), a dependency-report update (`docs/dependencies/*`),
and an IDE-config change (`.idea/misc.xml`).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.agents/memory/MEMORY.md
+++ b/.agents/memory/MEMORY.md
@@ -9,7 +9,7 @@ See [README.md](README.md) for the format and routing rules.
 
 ## Project (durable context & rationale)
 
-*(no entries yet)*
+- [codegen-coverage-blind-spot](project/codegen-coverage-blind-spot.md) — Why the `java`/`context` codegen modules report ~0% coverage — they run out-of-process in the forked Spine Compiler JVM, which JaCoCo/Kover do not instrument.
 
 ## Reference (external systems)
 
--- a/.agents/memory/project/codegen-coverage-blind-spot.md
+++ b/.agents/memory/project/codegen-coverage-blind-spot.md
@@ -1,0 +1,32 @@
+---
+name: codegen-coverage-blind-spot
+description: Why the `java`/`context` codegen modules report ~0% coverage — they run out-of-process in the forked Spine Compiler JVM, which JaCoCo/Kover do not instrument.
+metadata:
+  type: project
+  since: 2026-07-06
+---
+
+The `java` and `context` modules are Spine Compiler plugins/renderers (`JavaValidationPlugin`,
+the `*Generator` classes, the `expression/*` builders, the ProtoData plugin). They execute in the
+**forked `launch*SpineCompiler` (`JavaExec`) process** during the build, not in the `test` JVM.
+Kover/JaCoCo instrument only the test JVM, so this code reports **~0% line coverage** even though
+the `tests/*` and `context-tests` `.proto` fixtures exercise it on every build. The lone generator
+with real coverage — `UnsignedIntegerWarnings` (~82%) — has it because one spec instantiates it
+**in-process**. This is the main reason the repo total sits around ~12%.
+
+**Why:** So a future session seeing an alarmingly low total, or a codegen module at 0%, does not
+misread it as neglected/untested code or a quality regression and waste time re-deriving the cause.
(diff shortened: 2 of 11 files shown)


alexander-yevsyukov:
Addressed the dependency-report inconsistency noted in the Copilot review: the `docs/dependencies/` reports had been carried into the branch at `2.0.0-SNAPSHOT.447` from an earlier commit, while `version.gradle.kts` bumps to `.448`. Regenerated them for `.448` in c92dc9ddad — report headers and `pom.xml` are now consistent (0 remaining `.447` refs).

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

armiol (approved):

